PR #2, opened 2023-03-03
🔧(project) add missing gitlint configuration file

jmaupetit:
## Purpose

Our standard gitlint configuration is missing.

## Proposal

- [x] add missing `.gitlint` configuration file

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -1,0 +1,77 @@
+# All these sections are optional, edit this file as you like.
+[general]
+# Ignore certain rules, you can reference them by their id or by their full name
+# ignore=title-trailing-punctuation, T3
+
+# verbosity should be a value between 1 and 3, the commandline -v flags take precedence over this
+# verbosity = 2
+
+# By default gitlint will ignore merge commits. Set to 'false' to disable.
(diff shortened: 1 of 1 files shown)


.gitlint
```diff
@@ -0,0 +1,77 @@
...
+ignore=B6,UC1
+
+# [ignore-by-body]
+# Ignore certain rules for commits of which the body has a line that matches a regex
+# E.g. Match bodies that have a line that contains "release"
+# regex=(.*)release(.*)
+#
+# Ignore certain rules, you can reference them by their id or by their full name
+# Use 'all' to ignore all rules
+# ignore=T1,body-min-length
```
wilbrdt, line 77: You're missing the EOF carriage return
```suggestion
# ignore=T1,body-min-length

```

jmaupetit: :thinking: I don't think so. If it was the case, github would have displayed it as so. :wink:

wilbrdt: You're right, my bad!

wilbrdt (approved):
🎉

quitterie-lcs (approved):

PR #5, opened 2023-04-11
Add cunningham

jmaupetit:
## Purpose

We, OpenFUN, have our own design system called Cunningham. We need to use provided base components.

## Proposal

- [x] add base layout and integrate cunningham design system

During this work, we also improved the video `DailyViews` component:

- [x] stack videos views in a single chart


thwrk (approved):
Looks good to me!

PR #8, opened 2023-05-16
🐛 (project) fix `make bootstrap` in OSX (apple M1)

lebaudantoine:
## Purpose

We want `make bootstrap` instructions to run properly in OSX. 

## Proposal

Update `docker-compose` and `makefile` to support OSX profile. 

- [X] Add a new env variable `HOME` set to `/tmp` in `docker-compose` 
- [X] Replace `zcat` decompress tool by `gzcat` to decompress fixtures' archive in `makefile`.


CHANGELOG.md
```diff
@@ -13,5 +13,6 @@ and this project adheres to
 - Bootstrap base backend boilerplate
 - Implement video views endpoint
 - Bootstrap base frontend boilerplate using turborepo
+- Fix `make bootstrap` in OSX
```
jmaupetit, line 16: We usually don't update the CHANGELOG for changes that are related to the development environment and not the tool itself.

Makefile
```diff
@@ -49,7 +49,7 @@ bin/patch_statements_date.py:
 		-o bin/patch_statements_date.py
 
 data/statements.jsonl.gz:
-	mkdir -p data
+	rm -r data; mkdir -p data
```
jmaupetit, line 52: I would not do this in this rule, but rather in an explicit `clean` rule for example. Our goal here is to fetch the `statements.jsonl.gz` fixture, not delete the data directory content if it exists.

docker-compose.yml
```diff
@@ -31,6 +31,8 @@ services:
       - elasticsearch
 
   frontend:
+    environment:
+      HOME: /tmp
```
jmaupetit, line 35: Without being too pedantic, I would suggest to define the `environment` key after the image definition for this service. :wink:

Makefile
```diff
@@ -145,7 +145,7 @@ fixtures: \
 	curl -X PUT $(ES_URL)/$(ES_INDEX)/_settings \
 		-H 'Content-Type: application/json' \
 		-d '{"index": {"number_of_replicas": 0}}'
-	zcat data/statements.jsonl.gz | \
+	gzcat data/statements.jsonl.gz | \
```
jmaupetit, line 148: Have you tested this silly workaround instead? :sweat_smile: 

```suggestion
	zcat < data/statements.jsonl.gz | \
```

jmaupetit (commented):
Thank you for this! :pray:

Makefile
lebaudantoine: I haven't chosen this solution based on the [Vy32 own answer ](https://serverfault.com/a/570026) and prefer a more [recent one](https://serverfault.com/a/952618). I could definitely go for it.

> Apparently my answer is wrong. Linux zcat will read .zip and .Z files. The issue here isn't one of stdin vs. named file, it's that the Linux maintainers of the programs modified gzcat to read a large number of unrelated file types.

CHANGELOG.md
lebaudantoine: I have actually added it afterward to make the CI pass. My bad 😥

Makefile
```diff
@@ -48,8 +49,11 @@ bin/patch_statements_date.py:
 		https://raw.githubusercontent.com/openfun/potsie/v$(POTSIE_RELEASE)/scripts/patch_statements_date.py \
 		-o bin/patch_statements_date.py
 
+clean:
+	if [ -d $(WARREN_DATA_DIR) ]; then rm -r $(WARREN_DATA_DIR); fi
+
```
jmaupetit, line 54: Since this rule does not target a file, you should declare it as a `PHONY` one:

```suggestion
clean:
	if [ -d $(WARREN_DATA_DIR) ]; then rm -r $(WARREN_DATA_DIR); fi
.PHONY: clean
```

Makefile
lebaudantoine: I have created a proper `clean` rule, and added it in the `bootstrap` one. Opened to discuss if it's really relevant to do so, because we are running the `data/statements.jsonl.gz` rule in the fixture one, and I am not cleaning the data folder. It would introduce an inconsistent behavior.

Makefile
lebaudantoine: Got it ! I update it

Makefile
lebaudantoine: rule has been removed !

jmaupetit (approved):

PR #12, opened 2023-06-06
✨(backend) Add the LRS synchronous http backend

lebrunthibault:
## Purpose

We want warren to query a LRS instead of the current implementation using elasticsearch.
Ralph should not depend on a datastore like elastic but should consume data and do its aggregations using from a LRS

## Proposal

We should use the newly created LRSHTTP lrs client from the ralph-malph package backends
This will allow to query any LRS syncrhnously.
We will later add aggregations.

```python
from ralph.backends.http import LRSHTTP
```

- [x] install ralph-malph as a library
- [x] add a new backend using ralph's LRSHTTP (sync)
- [x] create a new video plugin using this backend
- [x] write tests
- [x] remove the previous elasticsearch backend


src/backend/plugins/video/warren_video/api.py
```diff
@@ -36,59 +42,40 @@ async def views(
...
+
+    for statement in statements:
+        # do not count video played less than the configured time
+        if (
+            statement["result"]["extensions"][RESULT_EXTENSION_TIME]
+            < settings.VIDEO_PLAYED_THRESHOLD_DURATION
+        ):
+            continue
+
+        date_key = statement["timestamp"][:10]
```
lebaudantoine, line 73: The timestamp of the statement might not match the explicit date format used by `daily_views_by_date` keys. To ensure consistency and remove the time component, it would be safer to parse the `timestamp` string using the arrow library and format it as `'YYYY-MM-DD'`.

Consider using:
```
arrow.get(statement["timestamp"]).format('YYYY-MM-DD')
```

Additionally, to avoid duplicated literals, it is recommended to extract the format `'YYYY-MM-DD'` into a shared constant at line 61 for better maintainability.

src/backend/plugins/video/warren_video/api.py
```diff
@@ -36,59 +42,40 @@ async def views(
...
+        if date_key in daily_views_by_date:
+            daily_views_by_date[date_key].views += 1
+
+    daily_views = list(daily_views_by_date.values())
+
+    return VideoViews(
+        total=sum(view.views for view in daily_views), daily_views=daily_views
```
lebaudantoine, line 80: nit: Have you considered adding `__add__` class methods to enable summation of `VideoViews` instances? This would enhance the usability and functionality of the class.

lebaudantoine: Overall, have you though using the `Counter` function from `Collections`. It might be a more `pythonic` way to count views. I haven't tested my code, however smth close to this should work : 

```
from collections import Counter

...

formatted_statements = [statement["timestamp"][:10] for statement in statements]
daily_views = [VideoDayViews(date=date, views=count) for date, count in Counter(formatted_statements).items()]
```
These onliners could also be split into relevant vars.
What do you think @jmaupetit ? 

Depending on the data, several list comprehensions might be faster than a unique list iteration.
We could profile this function, making sure my suggestion doesn't introduce performance regression.

Please note that my suggestion does not explicitly filter statements outside the date filters and also does not sort daily_views by dates. Some additional work will be required to address these aspects.

src/backend/plugins/video/warren_video/api.py
```diff
@@ -16,6 +20,8 @@
     prefix="/video",
 )
 
+logger = logging.getLogger("warren.api")
```
lebaudantoine, line 23: nit: `"warren.api"` literal is duplicated with `api/v1/__init__.py`

src/backend/plugins/video/__init__.py
```diff
@@ -0,0 +1 @@
+"""Warren Video plugin."""
```
lebaudantoine, line 1: nit: this description isn't redundant with the `warren_video/__init__.py`?

src/backend/core/warren/conf.py
```diff
@@ -18,11 +18,10 @@ class ESClientOptions(BaseModel):
...
 
-    # Elasticsearch
-    ES_HOSTS: Union[List[str], str] = None
-    ES_INDEX: str = "statements"
-    ES_INDEX_TIMESTAMP_FIELD: str = "timestamp"
-    ES_CLIENT_OPTIONS: ESClientOptions = ESClientOptions()
+    # LRS backend
+    LRS_HOSTS: Union[List[AnyHttpUrl], AnyHttpUrl] = "http://0.0.0.0:8100"
+    LRS_BASIC_USERNAME: str = "ralph"
+    LRS_BASIC_PASSWORD: str = "secret"
```
lebaudantoine, line 24: Just wanted to double-check, as a beginner back-end developer, whether these dummy/basic credentials need to be stored in a hidden environment file? Any guidance would be greatly appreciated!

src/backend/plugins/video/warren_video/api.py
```diff
@@ -36,59 +42,40 @@ async def views(
...
-            "daily_views": {
-                "date_histogram": {
-                    "field": settings.ES_INDEX_TIMESTAMP_FIELD,
-                    "calendar_interval": "day",
-                }
-            }
-        },
-        "size": 0,
+    query = {
+        "verb": PlayedVerb().id,
```
lebaudantoine, line 46: Hey @quitterie-lcs @jmaupetit, I've been thinking about something regarding the user experience during a watch session. If a user triggers the play/button multiple times, I'm concerned that it might be registering as multiple views in the end. Do you think it would be a good idea to implement some logic to aggregate these events? This way, we can ensure accurate view counts. What are your thoughts on this?

src/backend/plugins/video/warren_video/api.py
```diff
@@ -36,59 +42,40 @@ async def views(
...
+        logger.error("xAPI statements query failed: %s", error)
+        return VideoViews(total=0, daily_views=[])
+
+    daily_views_by_date = {
+        day.format("YYYY-MM-DD"): VideoDayViews(day=day.format("YYYY-MM-DD"))
+        for day in arrow.Arrow.range("day", filters.since, filters.until)
+    }
+
+    for statement in statements:
+        # do not count video played less than the configured time
```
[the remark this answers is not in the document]
lebaudantoine: Do we count only views happening at the beginning of a video as in [potsie l.189](https://github.com/openfun/potsie/blob/abba73265e8b479b0739bd3e340985e5f1679d03/src/dashboards/teachers/course.jsonnet#LL189C22-L189C42) ?

src/backend/core/warren/api/__init__.py
```diff
@@ -21,7 +20,6 @@ async def lifespan(app: FastAPI):
     yield
 
     # Properly shutdown the application
```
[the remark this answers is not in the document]
lebrunthibault: Yea, I hesitated about leaving this function here for later use but now we don't need to clean the es connection it's better to remove it altogether I agree.

src/backend/plugins/video/warren_video/api.py
```diff
@@ -36,59 +42,40 @@ async def views(
...
-    # Daily views buckets are supposed to be sorted by date range from the
-    # oldest to the newest record
-    idx = 0
-    for bucket in docs["aggregations"]["daily_views"]["buckets"]:
-        idx = video_views.daily_views.index(VideoDayViews(day=bucket["key"]), idx)
-        video_views.daily_views[idx].views = bucket["doc_count"]
-    return video_views
+    try:
+        statements = list(
+            lrs_client.read(target="/xAPI/statements/", query=LRSQuery(query=query))
```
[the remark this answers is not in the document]
lebrunthibault: Yes and it should : moved it to the settings

src/backend/core/warren/conf.py
lebrunthibault: I don't think this is an issue as the real credentials will sourced from the .env file (because they belong to a class extending pydantic BaseSettings). We could argue that it would be best to set no default for these test credentials but I left them as is to be coherent with ralph code

src/backend/plugins/video/warren_video/api.py
```diff
@@ -36,59 +42,44 @@ async def views(
...
+            day=day.format(settings.DATE_DAY_FORMAT)
+        )
+        for day in arrow.Arrow.range("day", filters.since, filters.until)
+    }
+
+    for statement in statements:
+        # Do not count video played less than the configured time
+        if (
+            statement["result"]["extensions"][RESULT_EXTENSION_TIME]
+            < settings.VIDEO_PLAYED_THRESHOLD_DURATION
```
jmaupetit, line 73: We want to ignore played events that occurred **after** the threshold:

```suggestion
        if (
            statement["result"]["extensions"][RESULT_EXTENSION_TIME]
            > settings.VIDEO_PLAYED_THRESHOLD_DURATION
```

src/backend/core/warren/conf.py
```diff
@@ -32,8 +33,10 @@ class Settings(BaseSettings):
     # API configuration
     MAX_DATETIMERANGE_SPAN: timedelta = timedelta(days=365)  # 1 year shift from since
     DEFAULT_DATETIMERANGE_SPAN: timedelta = timedelta(days=7)  # 7 days shift from until
+    VIDEO_PLAYED_THRESHOLD_DURATION: int = 30  # seconds
+    DATE_DAY_FORMAT: str = "YYYY-MM-DD"
```
jmaupetit, line 37: I think we could use the Django convention, e.g.:

```suggestion
    DATE_FORMAT: str = "YYYY-MM-DD"
```

WDYT?

jmaupetit (commented):
Smart move! :sparkles:

src/backend/core/warren/conf.py
```diff
@@ -18,11 +18,12 @@ class ESClientOptions(BaseModel):
...
 
-    # Elasticsearch
-    ES_HOSTS: Union[List[str], str] = None
-    ES_INDEX: str = "statements"
-    ES_INDEX_TIMESTAMP_FIELD: str = "timestamp"
-    ES_CLIENT_OPTIONS: ESClientOptions = ESClientOptions()
+    # LRS backend
+    LRS_HOSTS: Union[List[AnyHttpUrl], AnyHttpUrl] = "http://0.0.0.0:8100"
+    LRS_BASIC_USERNAME: str = "ralph"
+    LRS_BASIC_PASSWORD: str = "secret"
```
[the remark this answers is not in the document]
lebaudantoine: I agree.

src/backend/plugins/video/warren_video/api.py
lebrunthibault: I agree, maybe these suggestions should be tested by @cyrillay when he has a chance ? We could keep this pr simple for now and experiment with aggregations later on

Makefile
```diff
@@ -131,7 +131,7 @@ run-backend: ## run the backend server (development mode)
 	@$(COMPOSE) up -d backend
 	@echo "Waiting for backend to be up and running..."
 	@$(COMPOSE_RUN) dockerize -wait tcp://$(ES_COMPOSE_SERVICE):$(ES_PORT) -timeout 60s
-	@$(COMPOSE_RUN) dockerize -wait http://$(RALPH_COMPOSE_SERVICE):$(RALPH_RUNSERVER_PORT)/__heartbeat__ -timeout 60s
+	@$(COMPOSE_RUN) dockerize -wait tcp://$(RALPH_COMPOSE_SERVICE):$(RALPH_RUNSERVER_PORT)/__heartbeat__ -timeout 60s
```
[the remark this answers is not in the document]
lebrunthibault: definitely, I think this is right now

src/backend/core/warren/conf.py
lebrunthibault: sounds good to me !

src/backend/plugins/video/warren_video/api.py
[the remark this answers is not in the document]
lebrunthibault: actually it should be the same. I fixed this here : https://github.com/openfun/warren/pull/12/commits/eced4dce7c8aa1076489cf328ed2233fece44a5f

CHANGELOG.md
```diff
@@ -15,5 +15,9 @@ and this project adheres to
 - Bootstrap base backend boilerplate
 - Implement video views endpoint
 - Bootstrap base frontend boilerplate using turborepo
+- Run ralph in the docker compose
```
cyrillay, line 18: This was already the case, right ?

src/backend/plugins/video/warren_video/api.py
cyrillay: I agree that we can keep this PR simple for now, and optimize or refactor later on, because I believe the time taken by this code will be minor compared to the time taken to fetch the data from the LRS and its backend (for now).

cyrillay (approved):

src/backend/plugins/video/warren_video/api.py
cyrillay: If the code provided by @lebaudantoine works, I do think it would be great to have it, not because of performance (I think both codes would execute very quickly, so it's not worth profiling it IMO), but in terms of code readbility.

cyrillay: I would even turn the list comprehensions into a classic for loop, in order to add the cases you listed :
```
    Does not explicitly filter statements outside the date filters
    Does not sort daily_views by dates.
    Does not set to null dates without view.
```

In general, i'm not a big fan of list comprehensions, except when they are straightforward

lebaudantoine: What takes the most resources is generating the range of date, even for dates without any views. 
In general, list comprehensions are much faster than for loop, but I agree that they can hurt readability. 
Let's merge this PR as it is, no big deal.

lebaudantoine: I have quickly compared these two implementations, and profile them using `%timeit` ipython command.

(for my experience, I have generated a dummy list of statements following this tuple structure:
['YYYY-MM-DD', duration_value]

The duration values have been randomly generated using 'random' module, and are comprised between 0 and 1. 

_______

The actual workaround : 
```
def test_function(_statements, _start, _end):

    buffer = {
        day.format(DEFAULT_FORMAT): VideoDayViews(
            day=day.format(DEFAULT_FORMAT)
        )
        for day in arrow.Arrow.range("day", _start, _end)
    }
    
    for statement in _statements:

        if statement[1] >= 0.8:
            continue

        date_key = arrow.get(statement[0]).format(DEFAULT_FORMAT)
    
        if date_key in buffer:
            buffer[date_key].views += 1
    
    daily_views = list(buffer.values())
```
giving the following results :  
```
%timeit -r 100 test_function(statements, start, end)
> 110 ms ± 16.7 ms per loop (mean ± std. dev. of 100 runs, 10 loops each)
```

_______
The suggested implementation:
```
def test_function(_statements, _start, _end):
    
    range = arrow.Arrow.range("day", _start, _end)
        
    timestamps = [
        arrow.get(s[1])  # Convert timestamp to arrow date
        for s in _statements
        if s[1] < 0.8  # Filter on watch time duration
    ]

    counter = Counter(timestamps)
    daily_views = [VideoDayViews(day=date, views=counter[date]) for date in range]
```
giving the following results :  
```
%timeit -r 100 test_function(statements, start, end)
> 26.5 ms ± 1.21 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

lebrunthibault: sweet ! This looks nice, do you want to add a commit to the pr to implement this ? 

Isn't the part where we format to "YYYY-MM-DD" missing though ?

CHANGELOG.md
lebrunthibault: yes it was ! I think I forgot to add it in the previous merged pr where I add ralph to the docker compose :p

src/backend/plugins/video/warren_video/api.py
```diff
@@ -36,59 +44,46 @@ async def views(
     video_id: IRI, filters: Annotated[BaseQueryFilters, Depends()]
```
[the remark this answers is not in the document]
lebrunthibault: Yes ! This is the way fastapi does dependency injection in routes : https://fastapi.tiangolo.com/tutorial/dependencies/
This construct actually means the the BaseQueryFilters (a pydantic model) class will be instantiated and passer query parameters.
In the route we'll get the pydantic BaseQueryFilters object instantiated and we'll access the since and until fields. 
This type of setup allows us to validate the date range (using pydantic @root_validator) without adding any line or duplication in the routes :)

src/backend/plugins/video/warren_video/api.py
lebaudantoine: The date-to-string conversion was redundant : 

- Create a range of Arrow date object, format them to strings.
- Create an Arrow date from the timestamp, format it to a string.
- Instantiate a `VideoDayViews` with a string, which recreate an Arrow date object.

I use the Arrow date object as a key. I would appreciate extensive testing of this approach in the real function.

CHANGELOG.md
jmaupetit: BTW We'll squash/rewrite those for the first release.

src/backend/core/warren/conf.py
```diff
@@ -18,11 +18,11 @@ class ESClientOptions(BaseModel):
...
 
-    # Elasticsearch
-    ES_HOSTS: Union[List[str], str] = None
-    ES_INDEX: str = "statements"
-    ES_INDEX_TIMESTAMP_FIELD: str = "timestamp"
-    ES_CLIENT_OPTIONS: ESClientOptions = ESClientOptions()
+    # LRS backend
+    LRS_HOSTS: Union[List[AnyHttpUrl], AnyHttpUrl]
+    LRS_BASIC_USERNAME: str
+    LRS_BASIC_PASSWORD: str
```
jmaupetit, line 24: Maybe prefixing configuration variable related to authentication with `LRS_AUTH` would make sens. WDYT?

```suggestion
    LRS_AUTH_BASIC_USERNAME: str
    LRS_AUTH_BASIC_PASSWORD: str
```
By extension: `LRS_AUTH_OIDC_xxx`

src/backend/core/warren/conf.py
```diff
@@ -18,11 +18,11 @@ class ESClientOptions(BaseModel):
...
+    LRS_STATEMENTS_ENDPOINT: str = "/xAPI/statements"
```
jmaupetit, line 25: IMO `statements` is a  specified endpoint, hence, we should only configure the base URL for a given LRS host. WDYT?

```suggestion
    LRS_BASE_URL: str = "/xAPI"
```

I know it is related to Ralph's LRS client, not this project, but I would like your opinion on this. :wink:

lebrunthibault: If this is part of the lrs spect I agree, it's superflous to make it configurable

src/backend/plugins/video/warren_video/api.py
lebrunthibault: I see, but is this grouping by day or by datetime ? Didn't test it but the formatting was there to group by yyyy-mm-dd without taking into account the time.

lebaudantoine: I see, then my suggestion is irrelevant if statement timestamps contain time components. Performance might be on pair with your code.

lebaudantoine: The cognitive complexity of my suggestion might still be a bit simpler tho, I let judge, as you have more experience than me.

lebaudantoine (approved):

Issue #13, opened 2023-06-07
🧱(backend) Distributing warren docker images 


lebrunthibault:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
We should think about docker images distribution for warren so that teams can deploy warren without hassle.
We should consider the case of **plugins**. Should we have two separate docker images or just 2 different tags ? or maybe a tag per plugin ?

**Describe the solution you'd like**
The actual Dockerfile defines 2 stages : production with only the core and development that comprises core, plugins and dev dependencies.
We could add a 3rd stage named e.g. 'full' that comprises core and plugins without dev dependencies and users would be able to pull either fundocker/warren:core or fundocker/warren:full.
It's simple.

**Describe alternatives you've considered**
Please add your alternatives in the comments section :)
We could probably build instead two images, let's say fundocker/warren and fundocker/warren-full, as it's done with multiple fundocker images (marsha, funmooc ..). We could still use the same Dockerfile with multiple tag and just push two different images I suppose.

**Discovery, Documentation, Adoption, Migration Strategy**
Users will be able to `docker pull fundocker/warren:core` and `docker pull fundocker/warren:full`

**Do you want to work on it through a Pull Request?**
I could definitely work on this, let's see about that later.

jmaupetit:
Closing this as it has been implemented in #56

PR #14, opened 2023-06-07
Backend remove ci wrong docker hub publication job

lebrunthibault:
## Purpose

Remove ci wrong docker hub publication

The ci was configured to push the app image docker the docker hub under
a wrong name (ralph). I removed these steps and we will add
it back when we decide about which docker images scheme we use
(see https://github.com/openfun/warren/issues/13)

## Proposal

Remove the hub ci job

- [x] remove the hub circle ci job


wilbrdt (approved):
Fine with it!

PR #18, opened 2023-06-15
✨(project) add notebook service

cyrillay:
## Purpose

It would be nice to perform local adhoc operations (tests, benchmark…) on warren/LRS, in an easy way like notebooks.

## Proposal

This PR adds a jupyter notebook docker service that can be deployed along the other services with docker compose.

As an example, I added a simple notebook that benchmarks a request to the LRS, and a simple aggregation.


notebook/notebooks/benchmark_aggregation.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+```python
+import cProfile
+cProfile.run('query_and_aggregate()')
+```
+
+```python
+
+```
```
jmaupetit, line 77: Empty cell to remove :wink:

notebook/notebooks/benchmark_aggregation.md
```diff
@@ -0,0 +1,77 @@
...
+
+def query_and_aggregate():
+    df = query_lrs()
+    result = aggregate(df)
+    return result
+```
+
+```python
+import cProfile
+cProfile.run('query_and_aggregate()')
```
jmaupetit, line 72: FTR, maybe we should save this profile file?

```suggestion
cProfile.run('query_and_aggregate()', filename="benchmark-aggregation.profile")
```

jmaupetit (commented):
That's great!

Few suggestions:

1. I think you can squash your commits.
2. I would suggest the "`project`" scope for project/development related changes:

:sparkles:(project) add notebook service

3. Also: we don't add a new entry in the CHANGELOG for changes not related to the project from a user perspective. The CI job will fail, but we allow that.

Thanks :pray:

notebook/notebooks/benchmark_aggregation.md
cyrillay: Agreed. Do we want to version it as well ? I suggest I fix the version of ralph-malph used in the pip install of the first cell (`pip install ralph-malph==3.7.0`), so that we have something reproducible, even if we don't version the profiling file.

cyrillay:
@jmaupetit Thanks for the review ! I addressed your comments, let me know if there is anything missing

lebaudantoine (approved):
LGTM 🌟

lebaudantoine:
To access the warren package inside Jupyter Notebook, we need to update the volumes declared under the `docker-compose.yml file`. Currently, the package is not mounted inside the container, preventing access from within the Jupyter environment.

lebaudantoine:
> As you suggested during our call

dx is not optimal when finding the jupyter server urls among a lot of logs.

lebrunthibault (approved):

cyrillay:
> dx is not optimal when finding the jupyter server urls among a lot of logs.

I agree. I think for the moment I can at least add this to the documentation, WDYT ? 

> To access the warren package inside Jupyter Notebook, we need to update the volumes declared under the docker-compose.yml file. Currently, the package is not mounted inside the container, preventing access from within the Jupyter environment.

I agree. Do we want to make it available by default ? I can add the line for the volume

PR #21, opened 2023-06-19
✨(backend) implement video plugin endpoints

lebrunthibault:
## Purpose

Add complete_views, viewers and complete_viewers indicators to the video plugin and make them match their behavior in https://github.com/openfun/potsie

## Proposal

Implement 3 news video indicators for the video plugin:
- complete_views: the histogram of the completed statements on a video
- viewers: the total played statements on a video
- complete_viewers: the total completed statements on a video

The metric calculation has been isolated in two modules : count and histogram to be DRY.

- [x] Add the complete views indicator
- [x] Add the viewers indicator
- [x] Add the complete viewers indicator

### DIFF
--- a/src/backend/plugins/video/warren_video/viewers.py
+++ b/src/backend/plugins/video/warren_video/viewers.py
@@ -45,8 +45,6 @@ async def viewers(
     if verb == PlayedVerb():
         statements = filter(filter_played_views, statements)
 
-    unique_viewers = {
-        statement["actor"]["account"]["name"] for statement in statements
-    }
+    unique_viewers = {statement["actor"]["account"]["name"] for statement in statements}
 
-    return Count(total=len(list(unique_viewers)))
+    return Count(total=len(unique_viewers))


jmaupetit (commented):
Few general remarks/suggestions concerning project's layout:

- All tests for a plugin may be in a single `tests` directory at the root of the plugin directory so that we can easily ignore them while packaging the plugin.
- All API routes may be in a single `api` file / module at the root of the distributed plugin (the indicator semantic will be later integrated).

Some suggestions for endpoints patterns:

- unique views: `/{video_id:path}/views?unique=true`
- complete views: `/{video_id:path}/views/complete`
- complete viewers: `/{video_id:path}/views/complete?unique=true`

Alternative:

- viewers (unique views): `/{video_id:path}/views?unique=true`
- complete views: `/{video_id:path}/views?complete=true`
- complete viewers (unique complete views): `/{video_id:path}/views?complete=true&unique=true`

In this last case, it means that we only have a single API route for views. I think we don't need to stick to the xAPI semantic but rather focus on indicators semantic instead.

WDYT?

lebrunthibault:
> Few general remarks/suggestions concerning project's layout:
> 
> * All tests for a plugin may be in a single `tests` directory at the root of the plugin directory so that we can easily ignore them while packaging the plugin.
> * All API routes may be in a single `api` file / module at the root of the distributed plugin (the indicator semantic will be later integrated).
> 
> Some suggestions for endpoints patterns:
> 
> * unique views: `/{video_id:path}/views?unique=true`
> * complete views: `/{video_id:path}/views/complete`
> * complete viewers: `/{video_id:path}/views/complete?unique=true`
> 
> Alternative:
> 
> * viewers (unique views): `/{video_id:path}/views?unique=true`
> * complete views: `/{video_id:path}/views?complete=true`
> * complete viewers (unique complete views): `/{video_id:path}/views?complete=true&unique=true`
> 
> In this last case, it means that we only have a single API route for views. I think we don't need to stick to the xAPI semantic but rather focus on indicators semantic instead.
> 
> WDYT?



> Few general remarks/suggestions concerning project's layout:
> 
> * All tests for a plugin may be in a single `tests` directory at the root of the plugin directory so that we can easily ignore them while packaging the plugin.
> * All API routes may be in a single `api` file / module at the root of the distributed plugin (the indicator semantic will be later integrated).
> 
> Some suggestions for endpoints patterns:
> 
> * unique views: `/{video_id:path}/views?unique=true`
> * complete views: `/{video_id:path}/views/complete`
> * complete viewers: `/{video_id:path}/views/complete?unique=true`
> 
> Alternative:
> 
> * viewers (unique views): `/{video_id:path}/views?unique=true`
> * complete views: `/{video_id:path}/views?complete=true`
> * complete viewers (unique complete views): `/{video_id:path}/views?complete=true&unique=true`
> 
> In this last case, it means that we only have a single API route for views. I think we don't need to stick to the xAPI semantic but rather focus on indicators semantic instead.
> 
> WDYT?

Regarding declaration of a single api.py and tests/ directory I have a preference for grouping code by purpose instead of type. Would it be a fun convention to do otherwise ? Specifically for tests I think having them close to their source code has 2 benefits :
- makes it easier to understand what is tested and to update tests
- prevents the test code structure to loose sync with the source code structure when refactoring
I'm inspired on this by domain driven design.
Open to discussion on this though

Regarding the endpoints I see your point. We should be careful about pydantic response models though, I think it would be best to have each endpoint return a single type.
Still we could do :
- /views
- /views?complete=True (both histograms)
- /viewers
- /viewers?complete=True (both counts)

what do you think ?

lebrunthibault:
In https://github.com/openfun/warren/pull/21/commits/be590cb5178c9c3b73ed78ea99608399e150c440 I made the following corrections :
- moved the tests in a test folder outside of the warren_video package
- factored the 4 video endpoints into /views and /viewers with the completed boolean parameter

src/backend/plugins/video/warren_video/views.py
```diff
@@ -0,0 +1,48 @@
...
+            target="/xAPI/statements", query=LRSQuery(query=query)
+        )
+    except BackendException as error:
+        raise HTTPException(
+            status_code=status.HTTP_502_BAD_GATEWAY,
+            detail="xAPI statements query failed",
+        ) from error
+
+    if verb == PlayedVerb():
+        statements = filter(filter_played_views, statements)
```
lebaudantoine, line 46: I have some concerns and suggestions regarding the naming and functionality of `filter_played_views`:

1. The current function name doesn't provide a clear indication of its internal logic. It would be helpful if the name could be more descriptive to avoid checking the function definition (Especially that is a one-liner).

1. The term "played_views" in the function name implies that it only filters views that have been played, but in reality, it can accept any type of views (e.g., completed views) since there are no checks within the function. To improve clarity, we could either specialize the function or make the naming more generic.

I would suggest making this function a utility function for filtering videos based on their watching time. This can be achieved by adding a duration parameter that allows filtering based on a specified duration. Ex: 

```
def filter_views_by_duration(views, duration):
    return statement["result"]["extensions"][RESULT_EXTENSION_TIME] < duration
```

lebaudantoine: nit: have you check performance between filter vs list comprehension? As in the end, you convert your statements to a list. Might not be simple to implement, as there is a conditional filtering. Also please note, that this comment might be pre-mature optimization, feel free to ignore it.

src/backend/plugins/video/warren_video/viewers.py
```diff
@@ -0,0 +1,52 @@
...
+            target="/xAPI/statements", query=LRSQuery(query=query)
+        )
+    except BackendException as error:
+        raise HTTPException(
+            status_code=status.HTTP_502_BAD_GATEWAY,
+            detail="xAPI statements query failed",
+        ) from error
+
+    if verb == PlayedVerb():
+        statements = filter(filter_played_views, statements)
```
lebaudantoine, line 46: actually, all this logic is duplicated with the `views` endpoint and could be factorized. Please feel free to ignore my comment if you feel that's premature optimization. WDYT @cyrillay @jmaupetit ?

src/backend/plugins/video/warren_video/viewers.py
```diff
@@ -0,0 +1,52 @@
...
+
+    unique_viewers = set()
+    for statement in statements:
+        unique_viewers.add(statement["actor"]["account"]["name"])
```
lebaudantoine, line 50: for loop hurts performances in python 😬, we should avoid as much as possible using them IMO. I would suggest : 
```suggestion
    unique_viewers = { statement["actor"]["account"]["name"] for statement in statements }
```
Less lines 😊

src/backend/plugins/video/warren_video/viewers.py
```diff
@@ -0,0 +1,52 @@
...
+
+    return Count(total=len(list(unique_viewers)))
```
lebaudantoine, line 52: are you sure you need to convert a set to a list to get its length? This should work:
```suggestion
    return Count(total=len(unique_viewers))
```

lebaudantoine (commented):
I haven't reviewed the test files. I let you check my comment on your endpoints' implementation. Congrats 👏

src/backend/plugins/video/warren_video/metric/count.py
```diff
@@ -0,0 +1,14 @@
+"""Count metric for counting statements."""
```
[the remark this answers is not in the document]
lebrunthibault: yep !

src/backend/plugins/video/warren_video/metric/daytimeseries.py
```diff
@@ -0,0 +1,68 @@
+"""TimeSeries metric for generating day time series from statements."""
```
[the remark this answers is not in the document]
lebrunthibault: yes !

src/backend/plugins/video/warren_video/viewers.py
```diff
@@ -0,0 +1,52 @@
...
+    query = {
+        "verb": verb.id,
+        "activity": video_id,
+        "since": filters.since.isoformat(),
+        "until": filters.until.isoformat(),
+    }
+
+    try:
+        statements = lrs_client.read(
+            target="/xAPI/statements", query=LRSQuery(query=query)
```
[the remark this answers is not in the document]
lebrunthibault: yes this is also going to be refactored that's why I left it like this for now

src/backend/plugins/video/warren_video/views.py
```diff
@@ -0,0 +1,48 @@
...
+    query = {
+        "verb": verb.id,
+        "activity": video_id,
+        "since": filters.since.isoformat(),
+        "until": filters.until.isoformat(),
+    }
+
+    try:
+        statements = lrs_client.read(
+            target="/xAPI/statements", query=LRSQuery(query=query)
```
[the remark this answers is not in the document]
lebrunthibault: see above (will be refactored)

src/backend/plugins/video/warren_video/viewers.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+    except BackendException as error:
+        raise HTTPException(
+            status_code=status.HTTP_502_BAD_GATEWAY,
+            detail="xAPI statements query failed",
```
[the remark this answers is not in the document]
lebrunthibault: idem (will be refactored later)

src/backend/plugins/video/warren_video/viewers.py
lebrunthibault: This is not the same because the set() keeps only unique values :) I think these kind of improvements will be necessary but out of the scope of this pr for now

src/backend/plugins/video/warren_video/viewers.py
lebaudantoine: yes actually sorry for my comment, it's irrelevant. #29 is in charge of factorizing these lines. ! I resolve the conversation.

src/backend/plugins/video/warren_video/viewers.py
lebaudantoine: my code too, you can rewrite my suggestion as:
```
unique_viewers = set(statement["actor"]["account"]["name"] for statement in statements)
```
it should be the same.

lebaudantoine: Actually I haven't test it, I will.

src/backend/plugins/video/warren_video/views.py
lebrunthibault: I think these comments (especially the second) might be a bit premature as this is somewhat a work in progress.
Regarding your suggestion for the function it would mean we'll repeat `video_settings.VIEWS_COUNT_TIME_THRESHOLD` in the calls and I think it belongs semantically inside the function.

Regarding the name I see your point of view. I'm not sure a function should always describe its internals though and imo for a simple case like this it makes sense to "tag" it with its destination than a comprehensive name

src/backend/plugins/video/warren_video/viewers.py
```diff
@@ -0,0 +1,52 @@
...
+from warren_video.statement_filters import filter_played_views
+
+router = APIRouter()
+
+
+@router.get("/{video_id:path}/viewers")
+async def viewers(
+    video_id: IRI,
+    filters: Annotated[BaseQueryFilters, Depends()],
+    completed: bool = False,
```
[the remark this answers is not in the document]
lebrunthibault: this is julien's implementation ! The idea is to be dry when accessing and validating the daterange query filters. The completed query parameter is specific to certain routes so does not belong to the filters.

src/backend/plugins/video/warren_video/views.py
lebaudantoine: alright, in fact, after reviewing #29, this function has been refactored by @cyrillay. forgot my comments.

src/backend/plugins/video/warren_video/viewers.py
lebrunthibault: actually I misread your code both suggestions are valid ! I'll keep the first one which is best

PR #10, opened 2023-05-30
🏗️ (project) add Ralph LRS to the development stack

lebrunthibault:
## Purpose

We want warren to query ralph LRS as opposed to directly query the database backend (elasticsearch).
To do so, having ralph in the docker compose would be easier than running both projects locally.

## Proposal
Integrate fundocker/ralph into the docker-compose.yml

- [x] create the docker compose ralph service
- [x] configure the ralph service so that it queries the elastic setup in the elasticsearch service

PR #22, opened 2023-06-19
✨(backend) implement addition operation for VideoDayViews class

lebaudantoine:
## Purpose

Add new functionnalities on `VideoDayViews` class and prevent potentiel issue on `VideoViews`.

## Proposal

Added `__add__` and `__radd__` methods to `VideoDayViews` class for left-sided or right-sided addition operations

In the `VideoViews` class, there was a potential issue where the total and daily_views attributes could become desynchronized, causing bugs. To address this, `total` is now a read-only property that is dynamically computed based on the `daily_views` value.

Minor improvement: The current implementation recomputes the `total` property every time its value is accessed, which may impact performance. To mitigate this, I propose using the `@cached_property` decorator, which computes the value only once. However, it is important to ensure that the cached value of `total` is updated whenever `daily_views` changes. I would address this improvement in a next PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,5 +15,6 @@ and this project adheres to
 - Bootstrap base backend boilerplate
 - Implement video views endpoint
 - Bootstrap base frontend boilerplate using turborepo
+- Add addition on `VideoDayViews` class, prevent potential issue on `VideoViews`
 
 [unreleased]: https://github.com/openfun/warren
--- a/src/backend/plugins/video/warren_video/api.py
+++ b/src/backend/plugins/video/warren_video/api.py
@@ -35,8 +35,12 @@ class VideoDayViews(BaseModel):
 class VideoViews(BaseModel):
     """Model to represent video views."""
 
-    total: int
     daily_views: List[VideoDayViews]
+
+    @property
+    def total(self):
+        """Total of daily views."""
+        return sum(self.daily_views)
 
 
 @router.get("/{video_id:path}/views")
--- a/src/backend/plugins/video/warren_video/api.py
+++ b/src/backend/plugins/video/warren_video/api.py
@@ -23,6 +23,14 @@ class VideoDayViews(BaseModel):
     day: Date
     views: int = 0
 
+    def __add__(self, other):
+        """Emulates addition, left member."""
+        return self.views + other
+
+    def __radd__(self, other):
+        """Emulates addition, right member."""
+        return self.views + other
+
 
 class VideoViews(BaseModel):
     """Model to represent video views."""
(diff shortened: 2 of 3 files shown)


src/backend/plugins/video/warren_video/api.py
```diff
@@ -23,13 +23,25 @@ class VideoDayViews(BaseModel):
...
+        return self.views + other
+
 
 class VideoViews(BaseModel):
     """Model to represent video views."""
 
-    total: int
     daily_views: List[VideoDayViews]
 
+    @property
```
lebrunthibault, line 40: Attention : properties are not serialized by pydantic and so total would be removed from the api response.
We could keep the normal total field and set it in the __init__ method
With pydantic v2 we would be able to use computed fields

lebrunthibault (changes requested):

CHANGELOG.md
```diff
@@ -15,5 +15,6 @@ and this project adheres to
 - Bootstrap base backend boilerplate
 - Implement video views endpoint
 - Bootstrap base frontend boilerplate using turborepo
+- Add addition on `VideoDayViews` class, prevent potential issue on `VideoViews`
```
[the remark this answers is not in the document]
lebaudantoine: Actually it is, using the `+` before with `VideoDayViews` would trigger an error.

src/backend/plugins/video/warren_video/api.py
lebaudantoine: If we do what you suggest, we lose the interest of this PR:

lebaudantoine: Two quick workarounds :

CHANGELOG.md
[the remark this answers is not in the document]
lebrunthibault: I'm not sure but I think it's meant to be from the api user perspective point of view (see for example : https://github.com/openfun/warren/pull/10#discussion_r1211651520, to be confirmed)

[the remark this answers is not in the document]
lebaudantoine: I think, what @jmaupetit meant was, changes on dev setup, docker setup, etc.. are not meant to be tracked in the changelog.
These changes add functionalities to the warren package, thus for the user using the package. Let's quickly double-check during the standup.

src/backend/plugins/video/warren_video/api.py
lebaudantoine: What do you think about calling explicitly the property while returning the object @lebrunthibault?

lebrunthibault: This idea is interesting I think we should keep it. The dict override might not solve the problem because it will not override the json() method (used for json serialization I think).
pydantic v2 would be great but I understand its not yet ready for production user.

I would consider a solution of this type :

```python
class Histogram(BaseModel):
    """Histogram over a day range."""

    day_totals: List[HistogramDay]
    total: int

    def __init__(self, **data):
        """Instantiate a histogram from a list of day totals."""
        data["total"] = 0

        super().__init__(**data)

        self.total = sum(day.total for day in self.day_totals)
```

or a factory method `make_histogram(days: List[HistogramDays]) -> Histogram` would also be possible but would make it possible to instantiate an unsynced Histogram by itself

PR #27, opened 2023-06-20
👷(backend) Use astral-sh/ruff to make linting faster and simpler

### LINKED ISSUE #23: 👷(backend) Use astral-sh/ruff to make linting faster and simpler
## Feature Request

It would be nice to make the lint tasks in the ci and the Makefile be simpler and faster by introducing [ruff](https://github.com/astral-sh/ruff)

 Introduce [ruff](https://github.com/astral-sh/ruff) and use it to replace isort, flake8, pycodestyle and bandit.
we would keep ruff, black and mypy as [recommended by fastapi creator](https://twitter.com/tiangolo/status/1668026959668862977?s=20).

Check out the [ruff](https://github.com/astral-sh/ruff) readme. We'll use the `ruff check` command and configure the tool in *core/pyproject.toml*

Yes.

lebrunthibault:
## Purpose

Introduce ruff and use it to replace isort, flake8, pycodestyle and bandit by configuring it through pyproject.toml
This change might not be iso but should be close and brings a considerable speed improvement to the make lint command

## Proposal
Install ruff and activate checks to match at best previous linters
Do not activate all checks so we can still make ruff stricter.

Ruff current rule list can be found here : https://beta.ruff.rs/docs/rules/#pygrep-hooks-pgh

- [x] Add the astral-sh/ruff dev dependency 
- [x] Configure ruff so as to replace previous linters.
- [x] Add a lint-ruff-fix command to fix ruff errors automatically

closes #23


lebaudantoine (approved):

src/backend/core/pyproject.toml
```diff
@@ -107,5 +102,30 @@ testpaths = [
     "tests",
 ]
 
-[tool.bandit]
-exclude_dirs = ["tests",]
+[tool.ruff]
+fix = true
```
jmaupetit, line 106: I would prefer having a warning and use the CLI to fix it when I decide.

src/backend/core/pyproject.toml
```diff
@@ -107,5 +102,30 @@ testpaths = [
...
+]
+
+# Same as Black.
+line-length = 88
+
+# Assume Python 3.8.
+target-version = "py38"
+
+[tool.ruff.per-file-ignores]
+"*/tests/*" = ["S101"]
```
jmaupetit, line 127: Can you comment which rule you choose to ignore.

src/backend/core/pyproject.toml
```diff
@@ -107,5 +102,30 @@ testpaths = [
...
+# Enumerate all fixed violations.
+show-fixes = true
+
+select = [
+    "D",  # pydocstyle
+    "E",  # pycodestyle error
+    "F",  # Pyflakes F
+    "I",  # Isort
+    "S",  # bandit,
+    "W",  # pycodestyle warning
```
jmaupetit, line 117: I think we need to discuss this in comparison to popular projects using it. I think it's worth to add Pylint checks.

jmaupetit (commented):
Smart move!

Don't forget to remove the `src/backend/.pylintrc` configuration file!

src/backend/core/pyproject.toml
```diff
@@ -107,5 +102,30 @@ testpaths = [
...
+]
+
+# Same as Black.
+line-length = 88
```
wilbrdt, line 121:
```suggestion
```
We could probably ignore it if already supported by Black?

src/backend/core/pyproject.toml
lebrunthibault: - [fastapi](https://github.com/tiangolo/fastapi/blob/master/pyproject.toml) 
```
select = [
    "E",  # pycodestyle errors
    "W",  # pycodestyle warnings
    "F",  # pyflakes
    "I",  # isort
    "C",  # flake8-comprehensions
    "B",  # flake8-bugbear
]
ignore = [
    "E501",  # line too long, handled by black
    "B008",  # do not perform function calls in argument defaults
    "C901",  # too complex
]
```

- [airflow](https://github.com/apache/airflow/blob/main/pyproject.toml)
```
extend-select = [
    "I", # Missing required import (auto-fixable)
    "UP", # Pyupgrade
    "RUF100", # Unused noqa (auto-fixable)

    # implicit single-line string concatenation
    "ISC001",
    # We ignore more pydocstyle than we enable, so be more selective at what we enable
    "D101",
    "D106",
    "D2",
    "D3",
    "D400",
    # "D401", # Not enabled by ruff, but we don't want it
    "D402",
    "D403",
    "D412",
    "D419"
]
extend-ignore = [
    "D203",
    "D205",
    "D212",
    "D213",
    "D214",
    "D215",
    "E731",
]
```

[pandas](https://github.com/pandas-dev/pandas/blob/main/pyproject.toml)
```
select = [
  # pyflakes
  "F",
  # pycodestyle
  "E", "W",
  # flake8-2020
  "YTT",
  # flake8-bugbear
  "B",
  # flake8-quotes
  "Q",
  # flake8-debugger
  "T10",
  # flake8-gettext
  "INT",
  # pylint
  "PLC", "PLE", "PLR", "PLW",
  # misc lints
  "PIE",
  # flake8-pyi
  "PYI",
  # tidy imports
  "TID",
  # implicit string concatenation
  "ISC",
  # type-checking imports
  "TCH",
  # comprehensions
  "C4",
  # pygrep-hooks
  "PGH",
  # Ruff-specific rules
  "RUF",
]

ignore = [
  # space before : (needed for how black formats slicing)
  # "E203",  # not yet implemented
  # module level import not at top of file
  "E402",
  # do not assign a lambda expression, use a def
  "E731",
  # line break before binary operator
  # "W503",  # not yet implemented
  # line break after binary operator
  # "W504",  # not yet implemented
  # controversial
  "B006",
  # controversial
  "B007",
  # controversial
  "B008",
  # setattr is used to side-step mypy
  "B009",
  # getattr is used to side-step mypy
  "B010",
  # tests use assert False
  "B011",
  # tests use comparisons but not their returned value
  "B015",
  # false positives
  "B019",
  # Loop control variable overrides iterable it iterates
  "B020",
  # Function definition does not bind loop variable
  "B023",
  # Functions defined inside a loop must not use variables redefined in the loop
  # "B301",  # not yet implemented
  # Only works with python >=3.10
  "B905",
  # Too many arguments to function call
  "PLR0913",
  # Too many returns
  "PLR0911",
  # Too many branches
  "PLR0912",
  # Too many statements
  "PLR0915",
  # Redefined loop name
  "PLW2901",
  # Global statements are discouraged
  "PLW0603",
  # Docstrings should not be included in stubs
  "PYI021",
  # No builtin `eval()` allowed
  "PGH001",
  # compare-to-empty-string
  "PLC1901",
  # Use typing_extensions.TypeAlias for type aliases
  # "PYI026",  # not yet implemented
  # Use "collections.abc.*" instead of "typing.*" (PEP 585 syntax)
  # "PYI027",  # not yet implemented
  # while int | float can be shortened to float, the former is more explicit
  # "PYI041",  # not yet implemented

  # Additional checks that don't pass yet
  # Useless statement
  "B018",
  # Within an except clause, raise exceptions with ...
  "B904",
  # Magic number
  "PLR2004",
  # Consider `elif` instead of `else` then `if` to remove indentation level
  "PLR5501",
  # ambiguous-unicode-character-string
  "RUF001",
  # ambiguous-unicode-character-docstring
  "RUF002",
  # ambiguous-unicode-character-comment
  "RUF003",
  # collection-literal-concatenation
  "RUF005",
  # pairwise-over-zipped (>=PY310 only)
  "RUF007",
  # explicit-f-string-type-conversion
  "RUF010"
]
```

lebrunthibault: image

lebrunthibault: I propose we add B, C, RUF100 and ISC to be thorough ? That would make the config stronger than most projects probably but the tool is fast anyways

jmaupetit (approved):
Let's test it!

PR #29, opened 2023-06-21
✨(backend) Implement video views indicator

cyrillay:
## Purpose

This PR adds an indicator for getting the total number of views of a given video.

## Proposal

For now, the logic is simple and not many things are factored. I believe it is better to start with a naive approach, and factorize later on when we develop more indicators, so that we have a better idea on the common code that can be factorized, and how to do it.

Example return response : 
image


src/backend/plugins/video/warren_video/api.py
```diff
@@ -26,63 +17,18 @@
...
+    indicator = DailyVideoViews(
+        client=lrs_client,
+        activity=video_id,
+        since=filters.since.isoformat(),
+        until=filters.until.isoformat()
+    )
 
-    counter = Counter(timestamps)
-    daily_views = [VideoDayViews(day=date, views=counter[date]) for date in date_range]
+    computed_indicator = indicator.compute()
```
cyrillay, line 32: I removed the `try.. catch` block as it was already in the code of ralph when returning the statements. 
Thinking about it now, I think it would be nice to have a schema of the errors to return when an issue happened in the indicator.
```
Maybe we can return something like : 
{
status : "SUCCESS"
response : {...}
}
```
or 
```
{
status : "FAILED"
error_message : "The given video UID doesn't exist"
}
```

src/backend/plugins/video/warren_video/api.py
```diff
@@ -26,63 +17,18 @@
 logger = logging.getLogger(__name__)
 
 
-class VideoDayViews(BaseModel):
```
cyrillay, line 29: I will add the model again, I didn't notice the absence of it would cause the schema no longer to be shown in the swagger `/docs`

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,76 @@
...
+        # Group by day and calculate sum of events per day
+        count_by_date = filtered_view_duration.groupby('date')\
+            .count()\
+            .reset_index()\
+            .rename(columns={'id': 'count'})\
+            .loc[:, ['date', 'count']]
+        # Calculate the total number of events
+        total_count = len(filtered_view_duration.index)
+
+        return {"total_views": total_count, "views_count_by_date": count_by_date.to_dict('records')}
```
cyrillay, line 76: There are 2 options here 
`to_dict('list')`
```
{'date': [
  datetime.date(2023, 5, 25),
  datetime.date(2023, 5, 26),
  datetime.date(2023, 6, 21)
  ],
 'count': [
    7,
    5,
    8
]}
```
This avoids column repetition in the payload but needs a bit more logic to parse in the frontend

`.to_dict('records')`
```
[{'date': datetime.date(2023, 5, 25), 'count': 7},
 {'date': datetime.date(2023, 6, 20), 'count': 5},
 {'date': datetime.date(2023, 6, 21), 'count': 14}]
```

There is column repetition, but it is easier to parse in the frontend.
In my opinion, we should go with the second option until it proves to cause performance issues. (I did this option for another company and it worked well even with thousands of records. Here, for now, I don't see usecases where we will need to return so many records)

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,76 @@
...
+            "until": self.until,
+        }
+        return LRSQuery(query=query)
+
+    def get_statements(self) -> List:
+        filter_query = self.get_query()
+        statements = list(self.client.read(target=self.client.statements_endpoint, query=filter_query))
+        return statements
+
+    def compute(self) -> Dict:
```
cyrillay, line 49: I did a version with native python instead of pandas. I'm not sure which one is better, I am going to quickly profile them both and see which one is faster (my guess is native python): 

```
    def compute(self) -> Dict:
        """
        Queries the LRS backend and performs adequate filtering and aggregation to return
        the number of video views per day
        """
        raw_statements = self.get_statements()
        if not raw_statements:
            return {"total_views": 0, "views_count_by_date": {}}

        filtered_statements = []
        for statement in raw_statements:
            if statement['verb']['id'] == PlayedVerb().id:
                result_extensions = statement['result'].get('extensions', {})
                time = result_extensions.get(RESULT_EXTENSION_TIME, 0)
                if time < video_plugin_settings.VIEWS_COUNT_TIME_THRESHOLD:
                    filtered_statements.append(statement)

        count_by_date = {}
        for statement in filtered_statements:
            timestamp = statement['timestamp']
            date = arrow.get(timestamp).format('YYYY-MM-DD')
            if date not in count_by_date:
                count_by_date[date] = 0
            count_by_date[date] += 1

        total_count = len(filtered_statements)

        return {"total_views": total_count, "views_count_by_date": count_by_date}
```

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        raw_statements = self.get_statements()
+        if not raw_statements:
+            return {"total_views": 0, "views_count_by_date": {}}
+        flattened = pd.json_normalize(raw_statements)
+        # Filter out events whose verb is not "played"
+        filtered_verb = flattened[flattened['verb.id'] == PlayedVerb().id]
+        # Filter out videos played less than the configured time
+        filtered_view_duration = filtered_verb[filtered_verb[f'result.extensions.{RESULT_EXTENSION_TIME}']
+                                               < video_plugin_settings.VIEWS_COUNT_TIME_THRESHOLD].copy()
```
jmaupetit, line 62: Why not filtering in place instead to avoid too high memory usage?

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,76 @@
...
+        # Transform timestamp column into a day
+        filtered_view_duration['timestamp'] = pd.to_datetime(filtered_view_duration['timestamp'])
+        filtered_view_duration['date'] = filtered_view_duration['timestamp'].dt.strftime('%Y-%m-%d')
```
jmaupetit, line 65:
```suggestion
        filtered_view_duration['date'] = pd.to_datetime(filtered_view_duration['timestamp']).dt.strftime('%Y-%m-%d')
```

src/backend/plugins/video/warren_video/api.py
jmaupetit: Normalizing error messages is a good idea. Having a Pydantic model for this would be lovely. status/response would be more generic and preferable IMO.

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Group by day and calculate sum of events per day
+        count_by_date = filtered_view_duration.groupby('date')\
+            .count()\
+            .reset_index()\
+            .rename(columns={'id': 'count'})\
+            .loc[:, ['date', 'count']]
+        # Calculate the total number of events
+        total_count = len(filtered_view_duration.index)
```
jmaupetit, line 74: This needs to be battle tested.

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
jmaupetit: You should use the `VideoViews` model right?

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
cyrillay: Profiling results : 
Pandas : 
image

Vanilla python : 
image

I don't see any big difference. I think this benchmark should be made again with much more statements !

src/backend/plugins/video/warren_video/api.py
cyrillay: Yes, you are right, fields should have the same naming whether it failed or not.

src/backend/plugins/video/warren_video/base.py
```diff
@@ -0,0 +1,32 @@
+from abc import ABC, abstractmethod
```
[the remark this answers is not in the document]
cyrillay: I agree !

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,76 @@
...
+        in ISO format (YYYY-MM-DD or YYYY-MM-DDTHH:mm:ss.SSSZ")
+
+        Args:
+            client: The LRS backend from which the query is to be issued
+            activity: The UUID of the video on which to compute the metric
+            since: The beginning of the date range on which to compute the indicator
+            until: The end of the date range on which to compute the indicator
+        """
+        self.client = client
+        self.activity = activity
```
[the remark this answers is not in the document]
lebaudantoine: Agreed! One minor suggestion: in the documentation and the front-end project, we utilize uuid. To maintain consistency, perhaps video_uuid would be a better choice?

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
lebaudantoine: Out of curiosity, what is the `.loc` used for here?

lebaudantoine (commented):
Niiiceeee

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
cyrillay: Thanks for the insights on the two options. I think we can go with the second one for now, and eventually change later if it causes problems. Let me know if you agree.

Regarding the data validation on dates uniqueness, I don't think it is necessary, as per design of `.groupby('date')`, we should always have a unique set of dates. IMO, this validation would add a slight overhead without much benefit

lebaudantoine: I agree!

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
cyrillay: Well, I tried to do it, but I didn't manage to. In the end I got other issues that were harder to solve, or made the code look more complicated 🤷‍♂️  If you succeed, I would be curious to see how you did it !

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
cyrillay: It is for only keeping the columns that we want to return in the response

src/backend/plugins/video/warren_video/base.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @abstractmethod
+    def get_query(self) -> LRSQuery:
+        """
+        Returns the query that will be issued to the `client` to get the statements required to compute this indicator
+        """
+        pass
+
+    @abstractmethod
+    def get_statements(self) -> List:
```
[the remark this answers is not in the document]
cyrillay: I'm not familiar with abstract properties. When looking into the link, it looks like it is deprecated ?
I'm not sure I fully understand what is the benefit/drawback of one approach versus the other one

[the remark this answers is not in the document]
lebaudantoine: The `@abstractproperty` decorator is indeed deprecated, however the concept is not:
> the [property()](https://docs.python.org/3/library/functions.html#property) decorator is now correctly identified as abstract when applied to an abstract method:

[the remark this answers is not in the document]
lebaudantoine: Reminding my comment, it's nit-picking. Happy to discuss it.

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
lebaudantoine: Please find a colab with a [partial example](https://colab.research.google.com/drive/1ulD2cE8hIbpjGvxhw1q2ZokcfpJhy3ct?usp=sharing), lemme know if it helps you. Happy to help

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
cyrillay: I splitted the line into more steps so that it is easier to read. Let me know if that is better

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
cyrillay: You're right, I finally managed to do it. The grouped by worked fine but I had issue in the serialization when returning the API response, it always gave me `1970-01-01` or something like that. I managed to get it working now !

src/backend/core/warren/models/response.py
```diff
@@ -0,0 +1,22 @@
+"""Generic response class."""
+from enum import Enum
+from typing import Generic, TypeVar
+
+from pydantic.main import BaseModel
+
+
+class StatusEnum(str, Enum):
```
jmaupetit, line 8:
```suggestion
class StatusEnum(StrEnum):
```

src/backend/core/warren/models/response.py
```diff
@@ -0,0 +1,22 @@
...
+
+    SUCCESS = "SUCCESS"
+    FAILED = "FAILED"
+
+
+T = TypeVar("T")
+
+
+class Response(BaseModel, Generic[T]):
+    """Generic response class."""
```
jmaupetit, line 19:
```suggestion
    """Generic response model."""
```

src/backend/core/warren/models/response.py
```diff
@@ -0,0 +1,22 @@
...
+
+    status: StatusEnum
+    response: T
```
jmaupetit, line 22: WDYT?

```suggestion
    content: T
```

src/backend/core/warren/models/response.py
jmaupetit, line 1: Given the content of those models submodules, I think warren core models could be all defined in a single `models.py` module.

src/backend/plugins/video/warren_video/api.py
```diff
@@ -26,63 +19,23 @@
...
-        """Do not count video played less than the configured time."""
-        return (
-            statement["result"]["extensions"][RESULT_EXTENSION_TIME]
-            < video_settings.VIEWS_COUNT_TIME_THRESHOLD
+        return response
+    except KeyError as exception:
+        logger.error(exception)
+        return Response[Error](
+            status=StatusEnum.FAILED, response=Error(error_message=str(exception))
         )
```
jmaupetit, line 41: Try/except statement scope should be reduced at its smallest portion:

```suggestion
    except KeyError as exception:
        logger.error(exception)
        return Response[Error](
            status=StatusEnum.FAILED, response=Error(error_message=str(exception))
        )
    
    return response
```

src/backend/core/warren/base_indicator.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class BaseIndicator(ABC):
+    """Base class for an indicator.
+
+    Defines the signature of the methods that should be implemented
+    for indicators. Indicators should implement this interface.
+    """
+
+    client: BaseHTTP
```
jmaupetit, line 15: I think it's not required or used.

```suggestion
```

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,99 @@
...
+        client: The LRS backend from which the query is to be issued
+        video_uuid: The UUID of the video on which to compute the metric
+        date_range: The date range on which to compute the indicator. It has
+            2 fields, `since` and `until` which are dates or timestamps that must be
+            in ISO format (YYYY-MM-DD or YYYY-MM-DDTHH:mm:ss.SSSZ")
+    """
+
+    client: BaseHTTP
+    video_uuid: str
+    date_range: DatetimeRange
```
jmaupetit, line 32: We don't use them without instantiating the class, hence I would remove them.

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,99 @@
...
+        filter_query = self.get_lrs_query()
+        statements = list(
+            self.client.read(target=self.client.statements_endpoint, query=filter_query)
+        )
+        return statements
+
+    def compute(self) -> VideoViews:
+        """Computes the current indicator.
+
+        Gets the statements from the LRS, filters and aggregate to return the number of
```
jmaupetit, line 61:
```suggestion
        Gets the statements from the LRS, filters and aggregates them to return the number of
```

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,99 @@
...
+        video views per day.
+        """
+        response = VideoViews()
```
jmaupetit, line 64: I you keep in mind that this indicator will be used outside from an HTTP context (with the request/response semantic), I would rename this variable as `indicator`. WDYT?

```suggestion
        indicator = VideoViews()
```

jmaupetit (commented):
Wow, that's great! :muscle:

We are almost done. I only have few remarks.

src/backend/core/warren/models/response.py
cyrillay: Nice 👍

src/backend/core/warren/models/response.py
cyrillay: I agree, I prefer `content` too.

src/backend/plugins/video/warren_video/indicators/models.py
jmaupetit, line 1: I would move this file one level up (in the `warren_video` module).

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,96 @@
...
+        }
+        return LRSQuery(query=query)
+
+    def fetch_statements(self) -> List:
+        """Executes the LRS query to obtain statements required for this indicator."""
+        filter_query = self.get_lrs_query()
+        statements = list(
+            self.client.read(target=self.client.statements_endpoint, query=filter_query)
+        )
+        return statements
```
jmaupetit, line 51:
```suggestion
        return list(
            self.client.read(
                target=self.client.statements_endpoint, 
                query=self.get_lrs_query()
            )
        )
```

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def get_lrs_query(self) -> LRSQuery:
+        """Returns the LRS query for fetching required statements."""
+        query = {
+            "verb": PlayedVerb().id,
+            "activity": self.video_uuid,
+            "since": self.date_range.since.isoformat(),
+            "until": self.date_range.until.isoformat(),
+        }
+        return LRSQuery(query=query)
```
jmaupetit, line 43:
```suggestion
        return LRSQuery(
            query={
                "verb": PlayedVerb().id,
                "activity": self.video_uuid,
                "since": self.date_range.since.isoformat(),
                "until": self.date_range.until.isoformat(),
            }
        )
```

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,96 @@
...
+        Fetches the statements from the LRS, filters and aggregates them to return the
+        number of video views per day.
+        """
+        indicator = VideoViews()
+        raw_statements = self.fetch_statements()
+        if not raw_statements:
+            return indicator
+        flattened = pd.json_normalize(raw_statements)
+
+        # Filter out videos played less than the configured time
```
jmaupetit, line 65:
```suggestion
        # Filter out video played events before configured time threshold
```

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,96 @@
...
+    def fetch_statements(self) -> List:
+        """Executes the LRS query to obtain statements required for this indicator."""
+        filter_query = self.get_lrs_query()
+        statements = list(
+            self.client.read(target=self.client.statements_endpoint, query=filter_query)
+        )
+        return statements
+
+    def compute(self) -> VideoViews:
+        """Computes the current indicator.
```
jmaupetit, line 54: Maybe we can remove this generic statement?

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -0,0 +1,96 @@
...
+
+class DailyVideoViews(BaseIndicator):
+    """An indicator for the daily video views metric.
+
+    Args:
+        client: The LRS backend from which the query is to be issued
+        video_uuid: The UUID of the video on which to compute the metric
+        date_range: The date range on which to compute the indicator. It has
+            2 fields, `since` and `until` which are dates or timestamps that must be
+            in ISO format (YYYY-MM-DD or YYYY-MM-DDTHH:mm:ss.SSSZ")
```
jmaupetit, line 26: This should be documented in the constructor method instead.

jmaupetit (commented):
A final series of minor comments to fix and we'll be good to go. :pray:

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
jmaupetit, line 1: And I would also move this file in an `indicators.py` module.

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
cyrillay: I agree that it is too generic, but if I remove it, the rest of the description is too long to fit as a one linedocstring (rule E501). I modify it so that it is less generic, let me know if you have another idea ?

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
cyrillay: I am not sure I understand, the file is already in an `indicators` module ? You mean renaming this file as `indicators.py` ?

jmaupetit: I mean having a single `indicators.py` file with all indicators in there for now. IMO we don't need to have sub-modules (at least for now).

cyrillay: Okay, makes sense to me.

src/backend/core/warren/base_indicator.py
```diff
@@ -0,0 +1,25 @@
...
+from typing import List
+
+from ralph.backends.http.lrs import LRSQuery
+
+
+class BaseIndicator(ABC):
+    """Base class for an indicator.
+
+    Defines the signature of the methods that should be implemented
+    for indicators. Indicators should implement this interface.
```
lebaudantoine, line 12: nit: isn't it redundant?

src/backend/core/warren/base_indicator.py
```diff
@@ -0,0 +1,25 @@
...
+    def get_lrs_query(self) -> LRSQuery:
+        """Gets the LRS query for fetching statements."""
+
+    @abstractmethod
+    def fetch_statements(self) -> List:
+        """Executes the LRS query to obtain statements required for the indicator."""
+
+    @abstractmethod
+    def compute(self):
+        """Executes the query, and perform operations to get the indicator value."""
```
lebaudantoine, line 25: I like it, it's clearer 👏

lebaudantoine, line 25: nit, please feel free to ignore this comment:

```suggestion
    @abstractmethod
    def get_lrs_query(self) -> LRSQuery:
        """Gets the LRS query for fetching statements."""

    @abstractmethod
    def fetch_statements(self) -> List:
        """Executes the LRS query to obtain statements required for the indicator."""

    @abstractmethod
    def compute(self):
        """Executes the LRS query, and perform operations to get the indicator value."""
```

src/backend/core/warren/models.py
```diff
@@ -0,0 +1,28 @@
+"""Warren's core models."""
+from enum import Enum
+from typing import Generic, TypeVar
+
+from pydantic.main import BaseModel
+
+
+class StatusEnum(str, Enum):
```
lebaudantoine, line 8: As a junior, I wonder, is it a standard to add `Enum` in your class name? `class Status(str, Enum):` isn't explicit enough?

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        self.client = client
+        self.video_uuid = video_uuid
+        self.date_range = date_range
+
+    def get_lrs_query(self) -> LRSQuery:
+        """Returns the LRS query for fetching required statements."""
+        return LRSQuery(
+            query={
+                "verb": PlayedVerb().id,
```
lebaudantoine, line 40: After I reviewed #21, I wonder if we should increase modularity of this verb. @lebrunthibault is using a different verb to get completed views. WDYT @cyrillay?

src/backend/plugins/video/warren_video/models.py
```diff
@@ -0,0 +1,19 @@
...
+
+    date: Date
+    count: int = 0
+
+
+class VideoViews(BaseModel):
+    """Model to represent video views."""
+
+    total_views: int = 0
+    views_count_by_date: List[VideoDayViews] = []
```
lebaudantoine, line 19: @cyrillay I know it's not your design, and that we should not block this PR, so please feel free to ignore my comment. But, I feel this is a (small) bad design in a way, because these two attributes could be desynchronized. There are no mechanisms ensuring that the total updates with `views_count_by_date`. WDYT @jmaupetit ?

My initial proposal wasn't serializing properly. I would dig a second one with getter and setter.

lebaudantoine (approved):

src/backend/core/warren/base_indicator.py
cyrillay: Yes, definitely. I remove the second sentance

src/backend/core/warren/models.py
cyrillay: I didn't find anything showing that it was standard, but I encountered it a lot, and I think it good that it is explicit when you see it in the code, without having to go to the definition to understand it's an enum and not a classic constant

src/backend/plugins/video/warren_video/indicators.py
cyrillay: I think that it could be a good idea, I will probably do it when developing completed/unique views ;)

src/backend/plugins/video/warren_video/models.py
cyrillay: I think that adding a mechanism that avoids those two attributes to be desynchronized, is solving a problem that we don't have yet 🤷 I like the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle that talks about this

lebaudantoine: Alright, your call! I am pretty sure next time we'll use this class, we'll duplicate its instantiation logic. Anyway, let's keep it simple!

PR #31, opened 2023-06-26
👷(backend) Add mypy to the ci

### LINKED ISSUE #25: 👷(backend) Add mypy to the ci and correct violations in the codebase
## Feature Request

We could introduce mypy as a way to make our ci lint checks stronger and check and enforce strict typing in the codebase 

Install mypy in the devdependencies, add a lint-mypy makefile command and a step to the ci that runs mypy on the code

**Describe alternatives you've considered**
Some other type checkers exist and are popular like pyright or pyre but mypy is still the most standard choice.

Users will be able to `make lint-mypy` and mypy will run automatically in the ci

Yes

lebrunthibault:
## Purpose
Add mypy to the ci and correct violations in the codebase

We should make the typing experience stricter leveraging mypy and add it to the ci

## Proposal
Introduce mypy and lint the existing code

- [x] Install mypy as a dev dependency
- [x] Configure mypy so it understands the plugin architecture
- [x] Fix current violations to the code base
- [x] Make mypy work using make and in the ci 

closes #25

### DIFF
--- a/src/api/core/pyproject.toml
+++ b/src/api/core/pyproject.toml
@@ -112,9 +112,6 @@ convention = "google"
 [tool.mypy]
 mypy_path = "core,plugins/video"
 explicit_package_bases = "true"
-#plugins = [
-#  "pydantic.mypy"
-#]
 
 [[tool.mypy.overrides]]
 module = 'rfc3987.*'
--- a/src/api/core/pyproject.toml
+++ b/src/api/core/pyproject.toml
@@ -57,6 +57,7 @@ dev = [
     "pytest-cov==4.0.0",
     "pytest-httpx==0.22.0",
     "ruff==0.0.272",
+    "mypy==1.4.1",
 ]
 ci = [
     "twine==4.0.2",
@@ -107,3 +108,19 @@ target-version = "py38"
 [tool.ruff.pydocstyle]
 # Use Google-style docstrings.
 convention = "google"
+
+[tool.mypy]
+mypy_path = "core,plugins/video"
+explicit_package_bases = "true"
+#plugins = [
+#  "pydantic.mypy"
+#]
+
+[[tool.mypy.overrides]]
+module = 'rfc3987.*'
+ignore_missing_imports = "true"
+
+
+[[tool.mypy.overrides]]
+module = 'ralph.*'
+ignore_missing_imports = "true"
--- a/src/api/core/pyproject.toml
+++ b/src/api/core/pyproject.toml
@@ -54,6 +54,7 @@ dev = [
     "ipdb==0.13.11",
     "ipython==8.10.0",
     "mypy==1.4.1",
+    "pandas-stubs==2.0.2.230605",
     "pytest==7.2.1",
     "pytest-cov==4.0.0",
     "pytest-httpx==0.22.0",
@@ -110,8 +111,12 @@ target-version = "py38"
 convention = "google"
 
 [tool.mypy]
+exclude = [
+    '/tests/',
+]
 mypy_path = "core,plugins/video"
 explicit_package_bases = "true"
+
 
 [[tool.mypy.overrides]]
 module = 'rfc3987.*'
--- a/src/api/core/pyproject.toml
+++ b/src/api/core/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "fastapi==0.95.2",
     "importlib-metadata==6.6.0",
     "pandas==2.0.2",
-    "pydantic[dotenv]==1.10.8",
+    "pydantic[dotenv]==1.10.11",
     "rfc3987==1.3.8",
     "sentry-sdk[fastapi]==1.15.0",
     "uvicorn[standard]==0.20.0",
@@ -111,6 +111,9 @@ target-version = "py38"
 convention = "google"
 
 [tool.mypy]
+plugins = [
+  "pydantic.mypy"
+]
 exclude = [
     '/tests/',
 ]
--- a/src/api/core/pyproject.toml
+++ b/src/api/core/pyproject.toml
@@ -53,11 +53,11 @@ dev = [
     "httpx==0.24.1",
     "ipdb==0.13.11",
     "ipython==8.10.0",
+    "mypy==1.4.1",
     "pytest==7.2.1",
     "pytest-cov==4.0.0",
     "pytest-httpx==0.22.0",
     "ruff==0.0.272",
-    "mypy==1.4.1",
 ]
 ci = [
     "twine==4.0.2",
@@ -119,7 +119,6 @@ explicit_package_bases = "true"
 [[tool.mypy.overrides]]
 module = 'rfc3987.*'
 ignore_missing_imports = "true"
-
 
 [[tool.mypy.overrides]]
 module = 'ralph.*'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -171,6 +171,10 @@ jobs:
           name: Lint code with ruff
           command: ~/.local/bin/ruff --config core/pyproject.toml core plugins
           working_directory: src/api
+      - run:
+          name: Lint code with mypy
+          command: ~/.local/bin/mypy --config-file core/pyproject.toml core plugins
+          working_directory: src/api
   test-api:
     parameters:
       python-image:
--- a/Makefile
+++ b/Makefile
@@ -236,7 +236,8 @@ lint: \
 lint-api: ## lint api python sources
 lint-api: \
   lint-api-black \
-  lint-api-ruff
+  lint-api-ruff \
+  lint-api-mypy
 .PHONY: lint-api
 
 lint-api-black: ## lint api python sources with black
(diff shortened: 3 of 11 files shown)


lebrunthibault:
A way to make mypy understand the plugin architecture is to specify the different source code root folders in the `mypy_path` and set `explicit_package_bases`. 
This approach works but we need to manually maintain the mypy_path configuration for the next added plugins. 

Mypy will complain it cannot find library stubs as soon as we're importing anything from a plugin not added to the mypy path (even and especially from inside this plugin), so we'll not be able to ship untyped code anyway.
e.g. :
`plugins/video/warren_video/api.py:19: error: Cannot find implementation or library stub for module named "warren_video.conf"  [import]
`

lebrunthibault:
The construct used to setup the Date and DateTime fields does not play out well with mypy who cannot understand that these pydantic date types can take different inputs and be used transparently like python date datetime.

Apparently a [mypy plugin](https://docs.pydantic.dev/latest/mypy_plugin/) exists that might handle this kind of cases but I couldn't make it work locally with the current stable version of mypy (1.4.1) :
```bash
/usr/local/lib/python3.11/site-packages/pydantic/env_settings.py:22: error: INTERNAL ERROR -- Please try using mypy master on GitHub:                                                                              
https://mypy.readthedocs.io/en/stable/common_issues.html#using-a-development-mypy-build                                                                                                                            
If this issue continues with mypy master, please report a bug at https://github.com/python/mypy/issues                                                                                                             
version: 1.4.0                                                                                                                                                                                                     
/usr/local/lib/python3.11/site-packages/pydantic/env_settings.py:22: : note: please use --show-traceback to print a traceback when reporting a bug                                                                 
make: *** [Makefile:205: lint-mypy] Error 2 
```

For now I've used a two part solution that can be discussed:
- made Date and DateTime inherite date and datetime to silence mypy warnings on field access (e.g. date.days)
- manually silenced mypy with `#  type: ignore[arg-type]` on instantiation (e.g. ``VideoDayViews(day=day.format("YYYY-MM-DD"))

src/backend/core/pyproject.toml
```diff
@@ -57,6 +57,7 @@ dev = [
     "pytest-cov==4.0.0",
     "pytest-httpx==0.22.0",
     "ruff==0.0.272",
+    "mypy==1.4.1",
```
jmaupetit, line 60: Package list should be alphabetically sorted.

src/backend/core/pyproject.toml
```diff
@@ -107,3 +108,19 @@ target-version = "py38"
...
+explicit_package_bases = "true"
+#plugins = [
+#  "pydantic.mypy"
+#]
+
+[[tool.mypy.overrides]]
+module = 'rfc3987.*'
+ignore_missing_imports = "true"
+
+
```
jmaupetit, line 123:
```suggestion

```

jmaupetit (commented):
Thank you for this! I have the feeling that this integration will be a bit too hackish. Let's discuss this a bit more.

src/backend/core/warren/fields.py
```diff
@@ -6,7 +6,7 @@
 import rfc3987
 
 
-class Date:
+class Date(datetime.date):
```
quitterie-lcs, line 9:
```suggestion
class Date:
```

src/backend/core/warren/fields.py
```diff
@@ -32,7 +32,7 @@ def __modify_schema__(cls, field_schema):
         field_schema.update(type="string", example="2023-01-01")
 
 
-class Datetime:
+class Datetime(datetime.datetime):
```
quitterie-lcs, line 35:
```suggestion
class Datetime:
```

src/backend/core/warren/filters.py
```diff
@@ -11,8 +11,8 @@
 class DatetimeRange(BaseModel):
     """A date/time range."""
 
-    since: Datetime = None
-    until: Datetime = None
+    since: Datetime = None  # type: ignore[assignment]
```
quitterie-lcs, line 14:
```suggestion
    since: Datetime = None
```

src/backend/core/warren/filters.py
```diff
@@ -11,8 +11,8 @@
...
+    until: Datetime = None  # type: ignore[assignment]
```
quitterie-lcs, line 15:
```suggestion
    until: Datetime = None 
```

src/backend/core/warren/tests/test_fields.py
quitterie-lcs, line 1: It's not needed to have mypy on the tests as it is not distributed. A file pattern configuration is necessary to be added in the pyproject.toml

src/backend/plugins/video/tests/test_api.py
quitterie-lcs, line 1: Why is this test in this pull request ?

src/backend/core/pyproject.toml
```diff
@@ -108,3 +109,18 @@ target-version = "py38"
 [tool.ruff.pydocstyle]
 # Use Google-style docstrings.
 convention = "google"
+
+[tool.mypy]
+mypy_path = "core,plugins/video"
+explicit_package_bases = "true"
+#plugins = [
+#  "pydantic.mypy"
+#]
```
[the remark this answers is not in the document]
lebrunthibault: I tried unsuccesfully .. see comment above : https://github.com/openfun/warren/pull/31#issuecomment-1607658889

src/backend/core/warren/tests/test_fields.py
lebrunthibault: I would argue that mypy can be run on tests and that it would improve the typing coverage of the overral codebase, but that it can be indeed overkill, and make dev slower. [See this](https://calpaterson.com/mypy-hints.html#:~:text=Consider%20whether%20to%20include%20your%20tests).
But I'm ok with this I'll remove tests coverage for now :)

src/backend/core/warren/conf.py
```diff
@@ -3,16 +3,16 @@
...
 from pydantic import AnyHttpUrl, BaseModel, BaseSettings
 
 
 class ESClientOptions(BaseModel):
-    """Pydantic model for Elasticsearch additionnal client options."""
+    """Pydantic model for Elasticsearch additional client options."""
 
-    ca_certs: Path = None
-    verify_certs: bool = None
+    ca_certs: Optional[Path] = None
```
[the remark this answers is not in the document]
lebrunthibault: why not ? it makes the types correct and does not change the behavior of the model. (even with the existing behavior passing `ca_certs=None` is possible, and that's kind of counterintuitive imo)

src/backend/core/warren/filters.py
lebrunthibault: ditto

src/backend/core/pyproject.toml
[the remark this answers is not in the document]
lebrunthibault: made it work by bumping pydantic from 1.10.8 to to 1.10.11 !

lebrunthibault:
> The construct used to setup the Date and DateTime fields does not play out well with mypy who cannot understand that these pydantic date types can take different inputs and be used transparently like python date datetime.
> 
> Apparently a [mypy plugin](https://docs.pydantic.dev/1.10/mypy_plugin/) exists that might handle this kind of cases but I couldn't make it work locally with the current stable version of mypy (1.4.1) :
> 
> ```shell
> /usr/local/lib/python3.11/site-packages/pydantic/env_settings.py:22: error: INTERNAL ERROR -- Please try using mypy master on GitHub:                                                                              
> https://mypy.readthedocs.io/en/stable/common_issues.html#using-a-development-mypy-build                                                                                                                            
> If this issue continues with mypy master, please report a bug at https://github.com/python/mypy/issues                                                                                                             
> version: 1.4.0                                                                                                                                                                                                     
> /usr/local/lib/python3.11/site-packages/pydantic/env_settings.py:22: : note: please use --show-traceback to print a traceback when reporting a bug                                                                 
> make: *** [Makefile:205: lint-mypy] Error 2 
> ```
> 
> For now I've used a two part solution that can be discussed:
> 
> * made Date and DateTime inherit date and datetime to silence mypy warnings on field access (e.g. date.days) : [see diff](https://github.com/openfun/warren/pull/31/files#diff-c8e21394d1e36ebd94bf8c501d1fcca14f04f9bd5b3695637ee20d73f7c12f56L9-R9)
> * manually silenced mypy with `#  type: ignore[arg-type]` on instantiation : [see diff](https://github.com/openfun/warren/pull/31/files#diff-def7c87f928c05c103c3b3d328678832bf980ccea9519efc6819e1cec9536078L149-R150)
> 
> EDIT : made it work by bumping pydantic from 1.10.8 to 1.10.11 !

Actually, I don't think the plugin really solves our case.
The instantiation error is not there anymore but that's because we changed the way we do it.
The getters still throw errors (`.isoformat()`). I think we didn't setup the custom data types well as they should extend a type (see example in [pydantic doc](https://docs.pydantic.dev/1.10/usage/types/#custom-data-types)) both for mypy to understand how we use the objects and because objects returned are indeed considered to have this type

src/backend/core/warren/fields.py
lebrunthibault: ditto

src/api/core/pyproject.toml
```diff
@@ -107,3 +109,22 @@ target-version = "py38"
...
 convention = "google"
+
+[tool.mypy]
+plugins = [
+  "pydantic.mypy"
+]
+exclude = [
+    '/tests/',
+]
+mypy_path = "core,plugins/video"
```
lebaudantoine, line 120: Should have a more generic path than `plugins/video`?

lebrunthibault: I would have liked that but didn't find a way ! see this comment : https://github.com/openfun/warren/pull/31#issuecomment-1607636679

PR #32, opened 2023-06-26
Backend add git pre commit hook

### LINKED ISSUE #26: Add a pre commit hook to enforce strict checks at commit time instead of relying on the make commands and the ci
## Feature Request

Today we lint our code using manual make commands and automatic ci steps.
Like so it often happens to commit unlinted code that will fail in the pipeline, and it can be a waste of time.

To prevent this, we could setup a precommit hook that will automatically lint and fix our code at commit time.
It is especially relevant if we add ruff to the codebase (see https://github.com/openfun/warren/issues/23)
We could use black and ruff in the pre commit hook. Regarding mypy (if we implement it : see https://github.com/openfun/warren/issues/25), it can be done but we should decide if we want to endure the time penalty it takes on every commit. 

Pre commit hooks are standard git scripts but we should use the [pre commit library](https://pre-commit.com/) to make hooks manageable and versionable.

Yes

lebrunthibault:
## Purpose
Add a git pre commit hook so that some ci checks can be run locally on every commit

## Proposal
Create a bin/git-hook-pre-commit that is linked to the .git/hooks folder when running `make bootstrap`
and that executes black and ruff on every commit

Description...

- [x] setup the git pre commit hook during `make bootstrap`
- [x] check for trailing whitespace
- [x] run black in the hook
- [x] run ruff in the hook

closes #26

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,7 @@
 SHELL := /bin/bash
 
 # -- Docker
-# Get the current user ID to use for docker run and docker exec commands
-DOCKER_UID           = $(shell id -u)
-DOCKER_GID           = $(shell id -g)
-DOCKER_USER          = $(DOCKER_UID):$(DOCKER_GID)
-COMPOSE              = DOCKER_USER=$(DOCKER_USER) docker compose
+COMPOSE              = bin/compose
 COMPOSE_RUN          = $(COMPOSE) run --rm --no-deps
 COMPOSE_RUN_API      = $(COMPOSE_RUN) api
 COMPOSE_RUN_FRONTEND = $(COMPOSE_RUN) frontend
@@ -59,6 +55,14 @@ default: help
 
 .env:
 	cp .env.dist .env
+
+.git/hooks/pre-commit:
+	ln -sf ../../bin/git-hook-pre-commit .git/hooks/pre-commit
+
+git-hook-pre-commit:  ## Install git pre-commit hook
+git-hook-pre-commit: .git/hooks/pre-commit
+	@echo "Git pre-commit hook linked"
+.PHONY: git-hook-pre-commit
 
 .ralph/auth.json:
 	@$(COMPOSE_RUN) ralph ralph \
--- a/README.md
+++ b/README.md
@@ -41,6 +41,16 @@ We try to raise our code quality standards and expect contributors to follow
 the recommandations from our
 [handbook](https://handbook.openfun.fr).
 
+You can ensure your code is compliant by running the following commands:
+
+- `make lint` to run the linters
+- `make test` to run the tests
+
+Note that we also provide a git pre-commit hook to ease your life:
+```
+make git-hook-pre-commit
+```
+
 ## License
 
 This work is released under the MIT License (see [LICENSE](./LICENSE.md)).
--- a/bin/git-hook-pre-commit
+++ b/bin/git-hook-pre-commit
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+# Git pre-commit hook to lint the project code before committing.
+# This script is called by "git commit" with no arguments. The hook should
+# exit with a non-zero status and display an appropriate message if it
+# wants to stop the commit.
+
+# Usage:
+# This hook can be enabled by running `make git-hook-pre-commit`.
+
+# Get the commit reference against which we are comparing changes.
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+  against=HEAD
+else
+  # Initial commit: diff against an empty tree object
+  against=$(git hash-object -t tree /dev/null)
+fi
+
+# Check for trailing whitespace errors.
+echo "pre-commit hook: checking trailing whitespace…"
+git diff-index --check --cached "$against" --
+
+# Run the necessary commands to check modified code sources.
+if git diff --cached --name-only | grep -q 'src/api'; then
+  make lint-api
+fi
+if git diff --cached --name-only | grep -q 'src/app'; then
+  make lint-app
+fi
+if git diff --cached --name-only | grep -q 'src/frontend'; then
+  make lint-frontend
+fi
--- a/bin/compose
+++ b/bin/compose
@@ -3,4 +3,11 @@
 declare DOCKER_USER
 DOCKER_USER="$(id -u):$(id -g)"
 
-DOCKER_USER=${DOCKER_USER} docker compose "$@"
+extra_args=()
+# Prevent 'the input device is not a TTY' error from `docker compose run` by disabling TTY when standard output is not a TTY
+if [[ "$1" == run && "$(tty 2>/dev/null)" == "not a tty" ]]; then
+  extra_args=(run --no-TTY)
+  shift
+fi
+
+DOCKER_USER=${DOCKER_USER} docker compose "${extra_args[@]}" "$@"


Makefile
```diff
@@ -73,6 +78,7 @@ data/statements.jsonl.gz:
 bootstrap: ## bootstrap the project for development
 bootstrap: \
   .env \
+  git-hook-pre-commit \
```
lebaudantoine, line 81: Should we make it an opt-in feature for developers rather than enabling it by default? It might be better to let them choose whether to use it. 

Additionally, we can document this option in the `Contributing` section of the `Readme`. It's something that would facilitate contributions, but is not mandatory.

WDYT @jmaupetit ?

bin/git-hook-pre-commit
```diff
@@ -0,0 +1,62 @@
+#!/bin/sh
+#
+# Hook script to lint the backend code when it is about to be committed.
+# Called by "git commit" with no arguments.  The hook should
+# exit with non-zero status after issuing an appropriate message if
+# it wants to stop the commit.
+#
+# This hook should be enabled when doing `make bootstrap`.
```
lebaudantoine, line 8: ditto.

bin/git-hook-pre-commit
```diff
@@ -0,0 +1,62 @@
+#!/bin/sh
+#
+# Hook script to lint the backend code when it is about to be committed.
```
lebaudantoine, line 3: Your precommit hooks can be a bit misleading since they only cover a subset of the CI checks. This means that even if the precommit passes, the CI can still fail. Am I missing something?

Moreover, your solution is currently limited to the backend code. What about the front-end code or any potential future code sources, such as the LTI app?

Makefile
lebrunthibault: Good proposition ! We don't have a contribute file so I'll just remove it from the make boostrap for now  ^^

lebaudantoine: > we can document this option in the Contributing section of the Readme
I meant in the `Readme`

bin/git-hook-pre-commit
lebrunthibault: Good comment ! I've changed that to lint also the LTI app and the front end.

The script now runs the same checks as the ci for the 3 apps.
It doesnt run gitlint neither the changelog check because those can fail in the ci but we should still be able to commit

Makefile
lebrunthibault: done !

README.md
```diff
@@ -41,6 +41,15 @@ We try to raise our code quality standards and expect contributors to follow
 the recommandations from our
 [handbook](https://handbook.openfun.fr).
 
+You can ensure your code is compliant by running the following commands:
```
jmaupetit, line 44:
```suggestion
You can ensure your code is compliant by running the following commands:

```

README.md
```diff
@@ -41,6 +41,15 @@ We try to raise our code quality standards and expect contributors to follow
...
+- `make lint` to run the linters
+- `make test` to run the tests
+
+A git pre commit hook is available and can be installed using:
```
jmaupetit, line 48:
```suggestion
Note that we also provide a git pre-commit hook to ease your life:

```

Makefile
```diff
@@ -60,6 +56,11 @@ default: help
 .env:
 	cp .env.dist .env
 
+git-hook-pre-commit:
+	ln -s -f ../../bin/git-hook-pre-commit .git/hooks/pre-commit
+	@echo "Git pre-commit hook linked"
+.PHONY: git-hook-pre-commit
```
jmaupetit, line 62: Makefile rules  are usually file-path-based:

```suggestion
.git/hooks/pre-commit:
	ln -sf ../../bin/git-hook-pre-commit .git/hooks/pre-commit

git-hook-pre-commit:  ## Install git pre-commit hook
git-hook-pre-commit: .git/hooks/pre-commit
	@echo "Git pre-commit hook linked"
.PHONY: git-hook-pre-commit
```

Makefile
```diff
@@ -241,7 +242,7 @@ lint-api: \
 
 lint-api-black: ## lint api python sources with black
 	@echo 'lint-api:black started…'
-	@$(COMPOSE_RUN_API) black --config core/pyproject.toml core plugins
+	$(COMPOSE_RUN_API) black --config core/pyproject.toml core plugins
```
jmaupetit, line 245: Why echoing only this command?

bin/git-hook-pre-commit
```diff
@@ -0,0 +1,80 @@
...
+
+#set -e
+
+# If there are whitespace errors, print the offending file names and fail.
+echo "-----Checking trailing whitespace-----"
+git diff-index --check --cached "$against" --
+
+COMPOSE_RUN_API() {
+  docker compose run -T --rm --no-deps --quiet-pull api "$@"
+}
```
jmaupetit, line 67: dead code?

```suggestion
```

bin/git-hook-pre-commit
```diff
@@ -0,0 +1,80 @@
...
+  make lint-api; assert_command_successful "make lint-api"
+fi
+
+if git diff --cached --name-only | grep -q 'src/app'; then
+  make lint-app; assert_command_successful "make lint-app"
+fi
+
+if git diff --cached --name-only | grep -q 'src/frontend'; then
+  make lint-frontend; assert_command_successful "make lint-frontend"
+fi
```
jmaupetit, line 80: Not sure this is required as it's the git hook / shell responsibility.

```suggestion
if git diff --cached --name-only | grep -q 'src/api'; then
  make lint-api
fi

if git diff --cached --name-only | grep -q 'src/app'; then
  make lint-app
fi

if git diff --cached --name-only | grep -q 'src/frontend'; then
  make lint-frontend
fi
```

Makefile
```diff
@@ -2,11 +2,7 @@
 SHELL := /bin/bash
 
 # -- Docker
-# Get the current user ID to use for docker run and docker exec commands
-DOCKER_UID           = $(shell id -u)
-DOCKER_GID           = $(shell id -g)
-DOCKER_USER          = $(DOCKER_UID):$(DOCKER_GID)
-COMPOSE              = DOCKER_USER=$(DOCKER_USER) docker compose
+COMPOSE              = bin/compose
```
jmaupetit, line 5: :sunglasses:

bin/git-hook-pre-commit
lebaudantoine: In the current implementation, removing the assertions would lead to passing pre-commit checks, even if errors are present. Do you want me to dig it?

Makefile
lebaudantoine: That's actually a good question, I would guess that's a mistake. I fixed it.

lebaudantoine:
The phrase "Echoing -----Checking trailing whitespace----- in the hook feels a bit weird to me." is a bit awkward. Below are two examples of the hook's logs—one indicating success and the other showing an error:

```
➜  warren git:(backend-add-pre-commit-with-linting) ✗ git commit --fixup 45f8cf4
-----Checking trailing whitespace-----
[backend-add-pre-commit-with-linting 2572c8b] fixup! 🔨(backend) add a git pre commit hook
 1 file changed, 1 insertion(+), 1 deletion(-)
```

```
-----Checking trailing whitespace-----
lint-api:black started…
All done! ✨ 🍰 ✨
29 files left unchanged.
lint-api:ruff started…
plugins/video/tests/test_api.py:39:5: D202 [*] No blank lines allowed after function docstring (found 1)
plugins/video/tests/test_api.py:69:5: D202 [*] No blank lines allowed after function docstring (found 1)
Found 2 errors.
[*] 2 potentially fixable with the --fix option.
make: *** [lint-api-ruff] Error 1
```

WDYT @jmaupetit? 

Could we replace it with `pre-commit hook: checking trailing whitespace…`?

lebaudantoine:
Note: Providing a simple way to handle the removal of the pre-commit hook is essential for a smoother development workflow, especially in projects with multiple developers. WDYT? 

If we agree, I would add a make command.

jmaupetit:
> Note: Providing a simple way to handle the removal of the pre-commit hook is essential for a smoother development workflow, especially in projects with multiple developers. WDYT?
> 
> If we agree, I would add a make command.

You mean the removal of trailing spaces?

bin/git-hook-pre-commit
jmaupetit: :thinking: I think not: if a command fails, the command exit code should be > 0. This exit code should be catch by the git hook as a failure signal. I think @lebrunthibault's wrapper purpose is to echo a specific message, but it's not required IMO.

lebaudantoine: I have just re-tried, it seems that if the command fails, the exit code is not properly caught, because the commit still happens.

jmaupetit: One need to set `set -eo pipefail` on top of the script to ensure it stops with an appropriate error code when a command fails.

jmaupetit (approved):

PR #34, opened 2023-06-29
♻️(backend) refactor video plugin tests

cyrillay:
## Purpose

Currently, the tests of the video plugin are quite factorized, and they are not straightforward to understand for someone looking at the codebase for the first time.
Tests should be easy to read, in order to help to understand how the main functions work, and how to use them.
For those reasons, straightforward tests should facilitate external contributions, and simplify debugging and test refactoring during codebase development. 

Also, the test fixtures depend on the current date which can create un-wanted side effects, as they are not repeatable.

## Proposal

- Simplify the video plugin tests
- Reduce the level of factorization in the tests
- Remove the dependency to the current date in the tests
- Simplify the creation of the fixtures


src/backend/plugins/video/tests/test_api.py
```diff
@@ -1,148 +1,126 @@
...
-        path=lrs_client.statements_endpoint,
-        query=f"{query_string}&{urlencode({**params, 'limit': 500})}",
-        params="",
-        fragment="",
-    ).geturl()
-
-    httpx_mock.add_response(
-        url=lrs_url, method="GET", json={"statements": []}, status_code=200
+    response = await http_client.get(
+        url="/api/v1/video/uuid://fake-uuid/views",
```
lebrunthibault, line 53: didn't you remove some test cases by replacing the video_id parameter and hardcoding the video id in the url ?

src/backend/plugins/video/tests/test_api.py
```diff
@@ -1,148 +1,126 @@
...
+    view_2_time = datetime.fromisoformat(view_2_date).timestamp()
+    view_3_date = "2020-01-02T00:00:00"
+    view_3_time = datetime.fromisoformat(view_3_date).timestamp()
+    video_uuid = "uuid://ba4252ce-d042-43b0-92e8-f033f45612ee"
+    video_statements = [
+        VideoPlayedFactory.build(
+            [
+                {"object": {"id": video_uuid, "objectType": "Activity"}},
+                {"verb": {"id": PlayedVerb().id}},
+                {"result": {"extensions": {RESULT_EXTENSION_TIME: video_time}}},
```
lebrunthibault, line 83: isn't video_time a datetime here ? it's supposed to be a float I think

src/backend/plugins/video/tests/test_api.py
```diff
@@ -1,148 +1,126 @@
...
+    # Mock the call to the LRS so that it would return no statements, as it
+    # would do with no matching video
+    httpx_mock.add_response(
+        url=re.compile(r"^http://fake-lrs\.com/xAPI/statements\?.*$"),
+        method="GET",
+        json={"statements": []},
+        status_code=200,
     )
 
-    lrs_url = ParseResult(
```
lebrunthibault, line 77: this (too verbose ^^) construct test the exact url as opposed to your regex which is broader if I'm not mistaken, so the test could behave differently. This was inspired from [ralph tests](https://github.com/openfun/ralph/blob/ea9beda6602d7071392473fa9333405ba9867c0a/tests/backends/http/test_lrs.py#L236)

Still I'm ok with this more relaxed approach that make the tests clearer !

lebrunthibault (changes requested):

src/backend/plugins/video/tests/test_api.py
cyrillay: Yes, you are right. I add them again !

src/backend/plugins/video/tests/test_api.py
cyrillay: It is actually a float, because the `.timestamp()` method used to fill this variable, returns a UNIX timestamp as a float number (see screenshot below)
image

src/backend/plugins/video/tests/test_api.py
cyrillay: Yes, you are right : the regex makes it broader, and does only match the first part of the URL. But, as the mock is local to this test, and there is only one call made to the LRS, this shouldn't cause any issue, and I think it's a good compromise to make the tests clearer, and understand what's going on from someone reading the code.
It allows to focus on what is truely tested : not the LRS nor the call to the LRS, but the computation of the daily number of views

lebrunthibault (approved):

src/backend/plugins/video/tests/test_api.py
```diff
@@ -1,48 +1,36 @@
...
+    response = await http_client.get(
+        f"/api/v1/video/{video_id}/views", params=date_query_params
+    )
 
     assert response.status_code == 422
-    assert (
-        response.json().get("detail")[0].get("msg")
-        == f"'{parsed_video_id}' is not a valid 'IRI'."
-    )
+    assert "is not a valid 'IRI'." in response.json().get("detail")[0].get("msg")
```
jmaupetit, line 33: By not testing the parsed path, you've reduced the scope of this test :wink:

cyrillay: Is it a bad news ? For me it isn't 😅

src/backend/plugins/video/tests/test_api.py
cyrillay: Right. I can modify it so that it looks more natural (having the video time code between 0 and 1800 seconds (30 minutes))

src/backend/plugins/video/tests/test_api.py
cyrillay: For me, it was adding some complexity : adding a test case with a new video_id needs to adds a parsed video_id. One needs to know how URL parsing works, or do it in his python console to get the value he wants.

Also, `video_id` already contains parsed and unparsed routes, so we are still testing all the cases ? For me, this modification adds more laxity in the error message check, for the benefit of having a more concise, straightforward unit test, that anyone can understand and complete.

But I can add it back if you prefer !

jmaupetit: Ok then.

jmaupetit (approved):

PR #35, opened 2023-06-29
✨️(backend) add 2 endpoint parameters

cyrillay:
## Purpose

Add `unique` and `complete` endpoint parameters/routes for counting completed video events and/or unique viewers

- [x] Add logic in the indicator 
- [ ] Test the new logic
- [ ] Test various parameters combination

`/views` with `complete=False`
image

`/views` with `complete=True`
image

`/viewers` with `complete=False`
image

`/viewers` with `complete=True`
image

### DIFF
--- a/src/backend/core/warren/base_indicator.py
+++ b/src/backend/core/warren/base_indicator.py
@@ -1,8 +1,55 @@
 """Abstract class defining the interface for an indicator."""
+import hashlib
 from abc import ABC, abstractmethod
 from typing import List
 
+import pandas as pd
 from ralph.backends.http.lrs import LRSQuery
+
+from warren.conf import settings
+
+
+def parse_raw_statements(raw_statements) -> pd.DataFrame:
+    """Parse LRS statements, explode the columns, and add a `date` column."""
+    flattened = pd.json_normalize(raw_statements)
+    # Disable chained assignment warning to make the transformation inplace
+    pd.options.mode.chained_assignment = None
+    # Transform timestamp column into a date with day (YYYY-MM-DD)
+    flattened.loc[:, "date"] = pd.to_datetime(flattened.loc[:, "timestamp"]).dt.date
+    return flattened
+
+
+def add_actor_unique_id(statements: pd.DataFrame) -> pd.DataFrame:
+    """Add a `actor.uuid` column that uniquely identifies the agent.
+
+    Depending on the xAPI statements, the actor can be identified in 4 different ways :
+    https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#details-4. This
+    function handles the 4 cases and creates a `unique_actor_id` column that
+    can be used later without worrying about the 4 IFIs.
+    """
+    xapi_actor_identifier_columns = settings.XAPI_ACTOR_IDENTIFIER_PATHS.intersection(
+        set(statements.columns)
+    )
+
+    if len(xapi_actor_identifier_columns) == 0:
+        raise ValueError(
+            "There is no way of identifying the agent in submitted statements."
+        )
+    statements["actor.uuid"] = statements.apply(
+        lambda row: hashlib.sha256(
+            "-".join(str(row[col]) for col in xapi_actor_identifier_columns).encode()
+        ).hexdigest(),
+        axis=1,
+    )
+
+    return statements
+
+
+def pre_process_statements(statements: List) -> pd.DataFrame:
+    """Denormalize raw statements, and add utility columns."""
+    parsed = parse_raw_statements(statements)
+    with_unique_id = add_actor_unique_id(parsed)
+    return with_unique_id
 
 
 class BaseIndicator(ABC):
--- a/src/backend/core/warren/models.py
+++ b/src/backend/core/warren/models.py
@@ -1,6 +1,6 @@
 """Warren's core models."""
 from enum import Enum
-from typing import Generic, TypeVar
+from typing import Any, Dict, Generic, TypeVar
 
 from pydantic.main import BaseModel
 
@@ -26,3 +26,6 @@ class Response(BaseModel, Generic[T]):
 
     status: StatusEnum
     content: T
+
+
+XAPI_STATEMENT = Dict[str, Any]
--- a/src/backend/plugins/video/tests/test_api.py
+++ b/src/backend/plugins/video/tests/test_api.py
@@ -142,7 +142,7 @@ async def test_views_backend_query(
     video_views = (Response[VideoViews]).parse_obj(response.json()).content
 
     assert video_views["total_views"] == 1
-    assert video_views["views_count_by_date"][0] == {
+    assert video_views["count_by_date"][0] == {
         "date": datetime.utcnow().date().isoformat(),
         "count": 1,
     }
--- a/src/backend/plugins/video/warren_video/api.py
+++ b/src/backend/plugins/video/warren_video/api.py
@@ -7,7 +7,7 @@ from warren.backends import lrs_client
 from warren.fields import IRI
 from warren.filters import BaseQueryFilters, DatetimeRange
 from warren.models import Error, Response, StatusEnum
-from warren_video.indicators import DailyVideoViews
+from warren_video.indicators import DailyCompletedVideoViews, DailyVideoViews
 from warren_video.models import VideoViews
 
 router = APIRouter(
@@ -19,14 +19,23 @@ logger = logging.getLogger(__name__)
 
 @router.get("/{video_uuid:path}/views")
 async def views(
-    video_uuid: IRI, filters: Annotated[BaseQueryFilters, Depends()]
+    video_uuid: IRI,
+    filters: Annotated[BaseQueryFilters, Depends()],
+    complete: bool = False,
+    unique: bool = False,
 ) -> Response[VideoViews]:
     """Number of views for `video_uuid` in the `since` -> `until` date range."""
-    indicator = DailyVideoViews(
-        client=lrs_client,
-        video_uuid=video_uuid,
-        date_range=DatetimeRange(since=filters.since, until=filters.until),
-    )
+    indicator_kwargs = {
+        "client": lrs_client,
+        "video_uuid": video_uuid,
+        "date_range": DatetimeRange(since=filters.since, until=filters.until),
+        "is_unique": unique,
+    }
+
+    if complete:
+        indicator = DailyCompletedVideoViews(**indicator_kwargs)
+    else:
+        indicator = DailyVideoViews(**indicator_kwargs)
     try:
         response = Response[VideoViews](
             status=StatusEnum.SUCCESS, content=indicator.compute()
--- a/src/backend/plugins/video/warren_video/indicators.py
+++ b/src/backend/plugins/video/warren_video/indicators.py
@@ -2,12 +2,13 @@
 
 from typing import List
 
-import pandas as pd
 from ralph.backends.http.lrs import BaseHTTP, LRSQuery
 from ralph.models.xapi.concepts.constants.video import RESULT_EXTENSION_TIME
+from ralph.models.xapi.concepts.verbs.scorm_profile import CompletedVerb
 from ralph.models.xapi.concepts.verbs.video import PlayedVerb
-from warren.base_indicator import BaseIndicator
+from warren.base_indicator import BaseIndicator, pre_process_statements
 from warren.filters import DatetimeRange
+from warren.models import XAPI_STATEMENT
 from warren_video.conf import settings as video_plugin_settings
 from warren_video.models import VideoViews
 
@@ -19,19 +20,28 @@ class DailyVideoViews(BaseIndicator):
     a date range, the total number of views, and the number of views per day.
     """
 
-    def __init__(self, client: BaseHTTP, video_uuid: str, date_range: DatetimeRange):
+    def __init__(
+        self,
+        client: BaseHTTP,
+        video_uuid: str,
+        date_range: DatetimeRange,
+        is_unique: bool,
+    ):
         """Instantiate the indicator with its parameters.
 
         Args:
-            client: The LRS backend from which the query is to be issued
+            client: The LRS backend to query
             video_uuid: The UUID of the video on which to compute the metric
             date_range: The date range on which to compute the indicator. It has
                 2 fields, `since` and `until` which are dates or timestamps that must be
                 in ISO format (YYYY-MM-DD or YYYY-MM-DDTHH:mm:ss.SSSZ")
+            is_unique: If true, multiple views by the same actor are counted as
+                one
         """
         self.client = client
         self.video_uuid = video_uuid
         self.date_range = date_range
+        self.is_unique = is_unique
 
     def get_lrs_query(self) -> LRSQuery:
         """Returns the LRS query for fetching required statements."""
@@ -44,7 +54,7 @@ class DailyVideoViews(BaseIndicator):
             }
         )
 
-    def fetch_statements(self) -> List:
+    def fetch_statements(self) -> List[XAPI_STATEMENT]:
         """Executes the LRS query to obtain statements required for this indicator."""
         return list(
             self.client.read(
@@ -53,7 +63,7 @@ class DailyVideoViews(BaseIndicator):
         )
 
     def compute(self) -> VideoViews:
-        """Fetches statements and domputes the current indicator.
+        """Fetches statements and computes the current indicator.
 
         Fetches the statements from the LRS, filters and aggregates them to return the
         number of video views per day.
@@ -62,26 +72,24 @@ class DailyVideoViews(BaseIndicator):
         raw_statements = self.fetch_statements()
         if not raw_statements:
             return indicator
-        flattened = pd.json_normalize(raw_statements)
+        flattened = pre_process_statements(raw_statements)
 
-        # Filter out video played events before configured time threshold
+        # Filter out video played events before the configured time threshold
         def filter_view_duration(row):
             return (
                 row[f"result.extensions.{RESULT_EXTENSION_TIME}"]
                 >= video_plugin_settings.VIEWS_COUNT_TIME_THRESHOLD
             )
 
-        # Apply the filtering function to each row and update the DataFrame in place
+        # Apply the filtering function to each row
         filtered_view_duration = flattened[
             flattened.apply(filter_view_duration, axis=1)
         ]
 
-        # Disable chained assignment warning, as following Pandas advice doesn't work
-        pd.options.mode.chained_assignment = None
-        # Transform timestamp column into a date with day (YYYY-MM-DD)
-        filtered_view_duration.loc[:, "date"] = pd.to_datetime(
-            filtered_view_duration.loc[:, "timestamp"]
-        ).dt.date
+        # If the `is_unique` filter is selected, filter out rows with duplicate
+        # `actor.account.name`
+        if self.is_unique:
+            filtered_view_duration.drop_duplicates(subset="actor.uuid", inplace=True)
 
         # Group by day and calculate sum of events per day
         count_by_date = (
@@ -91,8 +99,91 @@ class DailyVideoViews(BaseIndicator):
             .rename(columns={"id": "count"})
             .loc[:, ["date", "count"]]
         )
+
         # Calculate the total number of events
         indicator.total_views = len(filtered_view_duration.index)
-        indicator.views_count_by_date = count_by_date.to_dict("records")
+        indicator.count_by_date = count_by_date.to_dict("records")
+
+        return indicator
+
+
+class DailyCompletedVideoViews(BaseIndicator):
+    """Daily Completed Video Views indicator.
+
+    For a given video, and a date range, this indicator computes the total number
+    of `completed` views events, and the number of views per day.
+    """
+
+    def __init__(
+        self,
+        client: BaseHTTP,
+        video_uuid: str,
+        date_range: DatetimeRange,
+        is_unique: bool,
+    ):
+        """Instantiate the indicator with its parameters.
+
+        Args:
+            client: The LRS backend from which the query is to be issued
+            video_uuid: The UUID of the video on which to compute the metric
+            date_range: The date range on which to compute the indicator. It has
+                2 fields, `since` and `until` which are dates or timestamps that must be
+                in ISO format (YYYY-MM-DD or YYYY-MM-DDTHH:mm:ss.SSSZ")
+            is_unique: If true, multiple views by the same actor are counted as
+                one
+        """
+        self.client = client
+        self.video_uuid = video_uuid
+        self.date_range = date_range
+        self.is_unique = is_unique
+
+    def get_lrs_query(self) -> LRSQuery:
+        """Returns the LRS query for fetching required statements."""
+        return LRSQuery(
+            query={
+                "verb": CompletedVerb().id,
+                "activity": self.video_uuid,
+                "since": self.date_range.since.isoformat(),
+                "until": self.date_range.until.isoformat(),
+            }
+        )
+
+    def fetch_statements(self) -> List[XAPI_STATEMENT]:
+        """Executes the LRS query to obtain statements required for this indicator."""
+        return list(
+            self.client.read(
+                target=self.client.statements_endpoint, query=self.get_lrs_query()
+            )
+        )
+
+    def compute(self) -> VideoViews:
+        """Fetches statements and computes the current indicator.
+
+        Fetches the statements from the LRS, filters and aggregates them to return the
+        number of video views per day.
+        """
+        indicator = VideoViews()
+        raw_statements = self.fetch_statements()
+        if not raw_statements:
+            return indicator
+        flattened = pre_process_statements(raw_statements)
+
+        # If the `is_unique` filter is selected, filter out rows with duplicate
+        # `actor.account.name`
+        if self.is_unique:
+            flattened.drop_duplicates(subset="actor.uuid", inplace=True)
+
+        # Group by day and calculate sum of events per day
+        count_by_date = (
+            flattened.groupby(flattened["date"])
+            .count()
+            .reset_index()
+            .rename(columns={"id": "count"})
+            .loc[:, ["date", "count"]]
+        )
+
+        # Calculate the total number of events
+        indicator.total_views = len(flattened.index)
+        indicator.count_by_date = count_by_date.to_dict("records")
 
         return indicator
--- a/src/backend/core/warren/conf.py
+++ b/src/backend/core/warren/conf.py
@@ -32,6 +32,13 @@ class Settings(BaseSettings):
     MAX_DATETIMERANGE_SPAN: timedelta = timedelta(days=365)  # 1 year shift from since
     DEFAULT_DATETIMERANGE_SPAN: timedelta = timedelta(days=7)  # 7 days shift from until
     DATE_FORMAT: str = "YYYY-MM-DD"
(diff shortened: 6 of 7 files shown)


src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -1,25 +1,32 @@
...
 from ralph.models.xapi.concepts.verbs.video import PlayedVerb
 from warren.base_indicator import BaseIndicator
 from warren.filters import DatetimeRange
 from warren_video.conf import settings as video_plugin_settings
-from warren_video.models import VideoViews
+from warren_video.models import DailyVideoViews
 
 
-class DailyVideoViews(BaseIndicator):
+class DailyVideoViewsIndicator(BaseIndicator):
```
jmaupetit, line 16: We already know that it's an indicator, it's not required to mention it as a suffix:

```suggestion
class DailyVideoViews(BaseIndicator):
```

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
```diff
@@ -1,25 +1,32 @@
...
     a date range, the total number of views, and the number of views per day.
     """
 
-    def __init__(self, client: BaseHTTP, video_uuid: str, date_range: DatetimeRange):
+    def __init__(
+        self,
+        client: BaseHTTP,
+        video_uuid: str,
+        date_range: DatetimeRange,
+        is_completed_views: bool,
```
jmaupetit, line 28:
```suggestion
        is_completed: bool,
```

src/backend/plugins/video/warren_video/indicators/daily_video_views.py
jmaupetit, line 1: As we still have few indicators, a single `indicators.py` file is preferred.

src/backend/core/pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     "elasticsearch[async]==8.6.2",
     "fastapi==0.95.2",
     "importlib-metadata==6.6.0",
+    "pandas==2.0.2",
```
cyrillay, line 35: Not sure why this appears, as it is already on main from the previous PR

src/backend/core/warren/base_indicator.py
```diff
@@ -0,0 +1,36 @@
+"""Abstract class defining the interface for an indicator."""
+from abc import ABC, abstractmethod
+from typing import List
+
+import pandas as pd
+from ralph.backends.http.lrs import LRSQuery
+
+
+def parse_raw_statements(raw_statements):
```
cyrillay, line 9: Should this be defined inside the `BaseIndicator` class ? I did it this way first, but pycharm gives me a warning that I don't use the `self`, and so can define the method outside of the class

src/backend/plugins/video/tests/test_api.py
```diff
@@ -141,16 +139,10 @@ async def test_views_backend_query(
 
     assert response.status_code == 200
 
-    video_views = VideoViews.parse_obj(response.json())
+    video_views = (Response[VideoViews]).parse_obj(response.json()).content
```
cyrillay, line 142: Those are changes from #34

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -0,0 +1,188 @@
...
+        # If the `is_unique_viewers` filter is selected, filter out rows with duplicate
+        # `actor.account.name`
+        if self.is_unique_viewers:
+            filtered_view_duration.drop_duplicates(
+                subset="actor.account.name", inplace=True
+            )
+
+        # Group by day and calculate sum of events per day
+        count_by_date = (
+            filtered_view_duration.groupby(filtered_view_duration["date"])
```
cyrillay, line 97: Should this be factorized ? IMO, it shouldn't, as it is quite specific to those two indicators (for now), and it would add complexity to add a function that would essentially be a wrapper of the pandas methods, without added value.

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -93,6 +102,87 @@ def filter_view_duration(row):
...
+    def fetch_statements(self) -> List:
+        """Executes the LRS query to obtain statements required for this indicator."""
+        return list(
+            self.client.read(
+                target=self.client.statements_endpoint, query=self.get_lrs_query()
+            )
+        )
+
+    def compute(self) -> VideoViews:
+        """Fetches statements and domputes the current indicator.
```
quitterie-lcs, line 160:
```suggestion
        """Fetches statements and computes the current indicator.
```

src/backend/core/warren/base_indicator.py
quitterie-lcs: Otherwise, you can define it as a static method in the `BaseIndicator` class (using the `@staticmethod` decorator)

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -0,0 +1,188 @@
...
+        Fetches the statements from the LRS, filters and aggregates them to return the
+        number of video views per day.
+        """
+        indicator = VideoViews()
+        raw_statements = self.fetch_statements()
+        if not raw_statements:
+            return indicator
+        flattened = parse_raw_statements(raw_statements)
+
+        # Filter out video played events before configured time threshold
```
lebrunthibault, line 76:
```suggestion
        # Filter out video played events before the configured time threshold
```

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -0,0 +1,188 @@
...
+        def filter_view_duration(row):
+            return (
+                row[f"result.extensions.{RESULT_EXTENSION_TIME}"]
+                >= video_plugin_settings.VIEWS_COUNT_TIME_THRESHOLD
+            )
+
+        # Apply the filtering function to each row and update the DataFrame in place
```
lebrunthibault: if the DataFrame is updated in place why is it assigned to `filter_view_duration` ?

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -0,0 +1,188 @@
...
+            filtered_view_duration.drop_duplicates(
+                subset="actor.account.name", inplace=True
+            )
+
+        # Group by day and calculate sum of events per day
+        count_by_date = (
+            filtered_view_duration.groupby(filtered_view_duration["date"])
+            .count()
+            .reset_index()
+            .rename(columns={"id": "count"})
```
lebrunthibault: this part could maybe be split into another python statement ?

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -0,0 +1,188 @@
...
+    def fetch_statements(self) -> List:
+        """Executes the LRS query to obtain statements required for this indicator."""
+        return list(
+            self.client.read(
+                target=self.client.statements_endpoint, query=self.get_lrs_query()
+            )
+        )
+
+    def compute(self) -> VideoViews:
+        """Fetches statements and domputes the current indicator.
```
lebrunthibault:
```suggestion
        """Fetches statements and computes the current indicator.
```

lebrunthibault (changes requested):

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -76,12 +85,12 @@ def filter_view_duration(row):
...
-        pd.options.mode.chained_assignment = None
-        # Transform timestamp column into a date with day (YYYY-MM-DD)
-        filtered_view_duration.loc[:, "date"] = pd.to_datetime(
-            filtered_view_duration.loc[:, "timestamp"]
-        ).dt.date
+        # If the `is_unique_viewers` filter is selected, filter out rows with duplicate
+        # `actor.account.name`
+        if self.is_unique_viewers:
+            filtered_view_duration.drop_duplicates(
+                subset="actor.account.name", inplace=True
```
wilbrdt, line 92: We would have to check for the four [valid IFI](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#2423-inverse-functional-identifier) an Actor can take:
- mbox
- mbox_sha1sum
- openid
- account

Is it planned for the future?

src/backend/plugins/video/warren_video/api.py
```diff
@@ -25,63 +17,35 @@
...
+    complete: bool = False,
+    unique: bool = False,
+) -> Response[VideoViews]:
+    """Number of views for `video_uuid` in the `since` -> `until` date range."""
+    if complete:
+        indicator = DailyCompletedVideoViews(
+            client=lrs_client,
+            video_uuid=video_uuid,
+            date_range=DatetimeRange(since=filters.since, until=filters.until),
+            is_unique_viewers=unique,
```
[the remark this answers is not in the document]
wilbrdt: Voting for is_unique_views!

wilbrdt (commented):
Alt text

src/backend/core/warren/base_indicator.py
cyrillay: Thanks quitterie, I wasn't aware of this decorator. 

I think I agree with you @lebrunthibault, I thought the function could belong to the class as a static method, but in the end, this helper could be useful outside of indicators too, so I'll leave it oustide the class for now. Later on, we could add this function and other helpers in a `helpers.py` file.

src/backend/core/warren/models.py
```diff
@@ -0,0 +1,28 @@
+"""Warren's core models."""
+from enum import Enum
+from typing import Generic, TypeVar
+
+from pydantic.main import BaseModel
+
+
+class StatusEnum(str, Enum):
+    """Enum for status types."""
```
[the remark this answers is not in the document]
cyrillay: The convention in python is to use `UPPER_CASE` for enum values, so I would prefer to keep them as is here (and eventually change them on ralph's side), see [here](https://docs.python.org/3/library/enum.html) and [here](https://stackoverflow.com/a/21359677) for reference

src/backend/plugins/video/warren_video/api.py
```diff
@@ -25,63 +17,35 @@
...
+        )
+    else:
+        indicator = DailyVideoViews(
```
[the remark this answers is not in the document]
cyrillay: In terms of code readability, I don't like instantiating classes using a variable holding the class name, and not directly the class itself, especially to save a few lines only. Also, it could hurt the linting, as we can't know before runtime what class is instantiated (so the IDE can't say what arguments should be passed)

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -0,0 +1,188 @@
...
+    Defines the daily video views indicator. It is, for a given video, and
+    a date range, the total number of views, and the number of views per day.
+    """
+
+    def __init__(
+        self,
+        client: BaseHTTP,
+        video_uuid: str,
+        date_range: DatetimeRange,
+        is_unique_viewers: bool,
```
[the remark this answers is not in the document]
cyrillay: Agreed !

src/backend/plugins/video/warren_video/indicators.py
cyrillay: You're right, it wasn't done in place actually. Fixed it

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -0,0 +1,188 @@
...
+        return LRSQuery(
+            query={
+                "verb": PlayedVerb().id,
+                "activity": self.video_uuid,
+                "since": self.date_range.since.isoformat(),
+                "until": self.date_range.until.isoformat(),
+            }
+        )
+
+    def fetch_statements(self) -> List:
```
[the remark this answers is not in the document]
cyrillay: Interesting options. In my opinion, I prefer the third option with a type alias : 
```
XAPI_STATEMENT = Dict[str, Any]
and -> List[XAPI_STATEMENT]:
``` 

The reason is that it gives a clear idea of what is returned by the function (xAPI statements), while not having too much constraints on the schema.
I am not in favor of the latter option : if we evolve the schema on the LRS side, we need to update it on warren as well, and synchronize the release of the LRS with the update of this repo. IMO, this cross-repo dependency can cause a big overhead.

Do you agree in selecting the type alias option ?

src/backend/plugins/video/warren_video/indicators.py
cyrillay: Interesting, I wasn't aware of this. If I understand correctly, we could have statements that don't even have the `actor.account` field, but `actor.mbox`/`actor.mbox_sha1sum`/`actor.openid` (depending on `actor.objectType`) ?
I might have missed it because the sample data 2.1K statements I use all have `actor.name`, it's a good catch !

src/backend/plugins/video/warren_video/indicators.py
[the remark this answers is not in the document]
lebrunthibault: Ok let's go for this !

src/backend/core/warren/base_indicator.py
jmaupetit: Using a class static method is a neat way to namespace a function, but it has to be related to the class. In this case a function looks perfectly fine to me. Maybe we will move its definition in an `utils.py`  module in a near future :man_shrugging:

src/backend/plugins/video/warren_video/api.py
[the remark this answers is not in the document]
jmaupetit: Or:

```python
indicator_kwargs = {
    "client": lrs_client,
    "video_uuid": video_uuid,
    "date_range": DatetimeRange(since=filters.since, until=filters.until),
    "is_unique_views": unique,
}


indicator = DailyCompletedVideoViews(**indicator_kwargs) if complete else DailyVideoViews(**indicator_kwargs)
```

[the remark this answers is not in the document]
cyrillay: The only downside is that we loose the IDE validation on the existence/non existence of the parameters for the class 🤷  But I don't have a strong opinion here, so I change it for the first option. (Not the second one that overlaps the 90 characters line limit, and I don't think it's worth adding parenthesis just for the sake of having the ternary expression)

src/backend/plugins/video/warren_video/indicators.py
cyrillay: I developed a `add_unique_actor_id` utility method that adds a column `actor.unique_actor_id` so that future developers don't have to worry about the 4 IRIs and 4 ways of identification. This column is added as part of the pre processing of the statements obtained (after normalizing), so future indicators can choose on how to use it (`df.drop_duplicates(subset="actor.unique_actor_id", inplace=True`). What do you think ?

Also, we don't have a `Statements` class, do we ? Or do you mean `src.ralph.api.models.LaxStatement` ?
I don't think it's a good idea to enforce strong typing here : end users might use different LRS and slightly different schemas with different validation rules. Also, we create (increase?) the dependency to ralph LRS, in the same mind of this [comment](https://github.com/openfun/warren/pull/35#discussion_r1254063827). What do you think ?

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -0,0 +1,188 @@
...
+        self,
+        client: BaseHTTP,
+        video_uuid: str,
+        date_range: DatetimeRange,
+        is_unique_viewers: bool,
+    ):
+        """Instantiate the indicator with its parameters.
+
+        Args:
+            client: The LRS backend from which the query is to be issued
```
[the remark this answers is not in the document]
lebrunthibault: what do you think ?

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -0,0 +1,188 @@
...
+            video_uuid: The UUID of the video on which to compute the metric
+            date_range: The date range on which to compute the indicator. It has
+                2 fields, `since` and `until` which are dates or timestamps that must be
+                in ISO format (YYYY-MM-DD or YYYY-MM-DDTHH:mm:ss.SSSZ")
+            is_unique_viewers: If true, multiple views by the same actor are counted as
+                one
+        """
+        self.client = client
```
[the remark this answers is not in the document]
lebrunthibault: what do you think ?

lebrunthibault:
🎉

src/backend/plugins/video/warren_video/indicators.py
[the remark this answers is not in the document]
cyrillay: Sorry I had missed this comment ! 
I didn't make them private because accessing or modifying them even after the class instantiation doesn't cause any problem. Users of the class could change the `is_unique` or `video_uuid` property and perform the computation again 🤷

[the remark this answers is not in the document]
lebrunthibault: I see ! I think it's simpler to recreate objects than mutate existing ones for the logical flow but I don't have a strong a priori on this, we can leave it like this

src/backend/plugins/video/warren_video/indicators.py
```diff
@@ -62,37 +73,119 @@ def compute(self) -> VideoViews:
...
+            flattened.drop_duplicates(subset="actor.unique_actor_id", inplace=True)
+
+        # Group by day and calculate sum of events per day
+        count_by_date = (
+            flattened.groupby(flattened["date"])
+            .count()
+            .reset_index()
+        )
+
+        renamed_df = count_by_date.rename(columns={"id": "count"}).loc[:, ["date", "count"]]
```
jmaupetit, line 186: Does the following work?

```suggestion
        count_by_date = (
            flattened.groupby(flattened["date"])
            .count()
            .reset_index()
            .rename(columns={"id": "count"}).loc[:, ["date", "count"]]
        )
```

src/backend/core/warren/base_indicator.py
```diff
@@ -2,9 +2,59 @@
...
+                return f"{row['actor.account.name']}-{row['actor.account.homePage']}"
+            else:
+                raise ValueError(
+                    "Multiple ways of identification for agent: more than one identification column is filled.")
+        else:
+            return row[non_nan_columns[0]]
+
+    statements['actor.unique_actor_id'] = statements.apply(generate_unique_agent_id, axis=1)
+
+    return statements
```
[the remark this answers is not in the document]
jmaupetit: Also, we can call this field `actor.uuid` if you store the hex representation of the checksum.

src/backend/plugins/video/warren_video/indicators.py
jmaupetit: It's a good idea to add an actor "UUID" column and write an utility for this :+1: 

Types in my example was a way to auto-document my suggestion: we don't need to be too pedantic on this :wink:

src/backend/plugins/video/warren_video/api.py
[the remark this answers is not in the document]
lebaudantoine: `kwargs` is a clever approach that enhances code readability by explicitly highlighting that only the class changes between the `if` and `else` statements, while the arguments remain unchanged. This improvement makes it easier to understand the logic and flow of the code.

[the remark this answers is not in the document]
lebaudantoine: `kwargs` would enhance code readability by explicitly highlighting that only the class changes between the if and else statements, while the arguments remain unchanged.

src/backend/core/warren/base_indicator.py
```diff
@@ -2,9 +2,59 @@
...
+
+
+def add_unique_actor_id(statements: pd.DataFrame) -> pd.DataFrame:
+    """Depending on the xAPI statements, the actor can be identified in 4 different ways :
+    https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#details-4
+
+    This function creates a `unique_agent_id` column that uniquely identifies the agent.
+    """
+
+    columns_to_check = ["actor.account.name", "actor.account.homePage", "actor.mbox", "actor.openid"]
```
lebaudantoine, line 28: shouldn't it a be constant set? 
```suggestion
    COLUMNS_TO_CHECK = {"actor.account.name", "actor.account.homePage", "actor.mbox", "actor.openid"}
```

src/backend/core/warren/base_indicator.py
```diff
@@ -2,9 +2,59 @@
...
+
+    for column in columns_to_check:
+        if column not in statements.columns:
+            statements[column] = np.nan
```
lebaudantoine, line 32:
```suggestion
    for missing_column in COLUMNS_TO_CHECK - set(statements.columns):
            statements[column] = np.nan
```

src/backend/core/warren/base_indicator.py
```diff
@@ -2,9 +2,59 @@
...
+        if len(non_nan_columns) == 0:
+            raise ValueError("No way of identifying the agent: all identification columns are NaN.")
+        elif len(non_nan_columns) > 1:
+            if "actor.account.name" in non_nan_columns and "actor.account.homePage" in non_nan_columns:
+                return f"{row['actor.account.name']}-{row['actor.account.homePage']}"
+            else:
+                raise ValueError(
+                    "Multiple ways of identification for agent: more than one identification column is filled.")
+        else:
+            return row[non_nan_columns[0]]
```
lebaudantoine, line 45: Overall: I find unclear adding nan column, and then checking which columns are not nan. I would be more than happy to discuss it. Why not using set diff?

`non_nan_columns = COLUMNS_TO_CHECK - set(row.index)`

lebaudantoine: This need to be tested, but some suggestions can be reused. I find it clearer to avoid nested `if`/`else`, it reduces cognitive complexity.

```suggestion
        non_nan_columns = COLUMNS_TO_CHECK - set(row.index)
        
        if not non_nan_columns:
                raise ValueError("No way of identifying the agent: all identification columns are NaN.")
        
        if len(non_nan_columns) == 1:
                return row[non_nan_columns[0]]

        if not {"actor.account.name", "actor.account.homePage"}.issubset(non_nan_columns):
                raise ValueError("Multiple ways of identification for agent: more than one identification column is filled.")

        return f"{row['actor.account.name']}-{row['actor.account.homePage']}"
```

src/backend/core/warren/base_indicator.py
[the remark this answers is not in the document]
cyrillay: Yes, that can work too. The only thing is I would need to do it before the parsing (because currently i denormalize the json as soon as possible to work with it, so the `actor` field result in multiple fields)

Also, we have something human readable that can still eventually be mapped to the original statements (I could improve the format of the `'actor.unique_actor_id'` so that we can know how it was identified before the creation of this column). In the `'actor.unique_actor_id'` field I could eventually store a JSON string of the original `actor` object ?

If we do the checksum, we can’t retrieve the original actor easily (not a huge deal)

WDYT ?

src/backend/plugins/video/warren_video/indicators.py
cyrillay: Yes, it does, I had changed it because of @lebrunthibault [other comment](https://github.com/openfun/warren/pull/35#discussion_r1253087211) 😰 but I prefer to chain the operations so I will change it

src/backend/core/warren/base_indicator.py
cyrillay: Okay, personally I find it harder to read and less natural, but I modify it

src/backend/core/warren/base_indicator.py
```diff
@@ -2,9 +2,59 @@
...
+    flattened = pd.json_normalize(raw_statements)
+    # Disable chained assignment warning to make the transformation inplace
+    pd.options.mode.chained_assignment = None
+    # Transform timestamp column into a date with day (YYYY-MM-DD)
+    flattened.loc[:, "date"] = pd.to_datetime(flattened.loc[:, "timestamp"]).dt.date
+    df = add_unique_actor_id(flattened)
+    return df
+
+
+def add_unique_actor_id(statements: pd.DataFrame) -> pd.DataFrame:
```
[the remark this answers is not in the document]
cyrillay: It's true that we uniform it, but I wanted to make it clear that we're adding a column here. Maybe `add_uniformized_actor_id` ?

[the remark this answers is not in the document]
lebaudantoine: True that my suggestion is missing the addition part, my bad. If you feel that `add_uniformized_actor_id` is fine let's go, however feel free to stick to your original naming 😁

src/backend/core/warren/base_indicator.py
cyrillay: Yes, I agree. Pycharm gives a warning when defining a constant inside a function, but I think we should ignore it and follow the convention.

src/backend/core/warren/base_indicator.py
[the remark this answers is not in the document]
cyrillay: No no, naming is important and the last one is definitely better ;)

src/backend/core/warren/base_indicator.py
```diff
@@ -2,9 +2,75 @@
...
+    function handles the 4 cases and creates a `unique_actor_id` column that
+    can be used later without worrying about the 4 IFIs.
+    """
+    COLUMNS_TO_CHECK = {
+        "actor.account.name",
+        "actor.account.homePage",
+        "actor.mbox",
+        "actor.mbox_sha1sum",
+        "actor.openid",
+    }
```
jmaupetit, line 34: I think we can make this a global `XAPI_ACTOR_IDENTIFIER_PATHS` configuration variable.

src/backend/core/warren/base_indicator.py
```diff
@@ -2,9 +2,75 @@
...
+                raise ValueError(
+                    "Multiple ways of identification for agent: "
+                    "more than one identification column is filled."
+                )
+        else:
+            raise ValueError("Unexpected number of identification columns.")
+
+    statements["actor.unique_actor_id"] = statements.apply(
+        generate_unique_agent_id, axis=1
+    )
```
jmaupetit, line 62: I haven't tested it, but WDYT about the following alternative implementation?

```suggestion
    xapi_actor_identifier_columns = XAPI_ACTOR_IDENTIFIER_PATHS.intersection(set(statements.columns))
    
    if len(xapi_actor_identifier_columns) == 0:
        raise ValueError(
            "There is no way of identifying the agent in submitted statements."
        )

    statements["actor.uid"] = statements.apply(
        lambda row, columns: hashlib.sha256("-".join(row[columns]).hexdigest(), 
        axis=1, 
        xapi_actor_identifier_columns
    )
```

src/backend/core/warren/base_indicator.py
```diff
@@ -2,9 +2,75 @@
...
+    """Parse LRS statements, explode the columns, and add a `date` column."""
+    flattened = pd.json_normalize(raw_statements)
+    # Disable chained assignment warning to make the transformation inplace
+    pd.options.mode.chained_assignment = None
+    # Transform timestamp column into a date with day (YYYY-MM-DD)
+    flattened.loc[:, "date"] = pd.to_datetime(flattened.loc[:, "timestamp"]).dt.date
+    return flattened
+
+
+def add_uniformized_actor_id(statements: pd.DataFrame) -> pd.DataFrame:
```
jmaupetit, line 20: I would prefer the following:

```suggestion
def add_actor_unique_id(statements: pd.DataFrame) -> pd.DataFrame:
```

src/backend/core/warren/base_indicator.py
cyrillay: Neat 🔥

jmaupetit (commented):
I only have one minor remark :wink:

jmaupetit (approved):

PR #36, opened 2023-07-18
(backend) Add base indicator tests

cyrillay:
## Purpose

Base indicators was merged before it had tests, in order to limit the size of the PR, and as it already had many discussions. This PR adds the missing tests.

## Proposal

Add unit tests for the `parse_raw_statements` and the `add_actor_unique_id` functions of the `base_indicator.py` file.

### DIFF
--- a/src/backend/core/warren/tests/test_base_indicator.py
+++ b/src/backend/core/warren/tests/test_base_indicator.py
@@ -1,0 +1,95 @@
+"""Test the functions from the BaseIndicator class."""
+import pandas as pd
+
+from warren.base_indicator import add_actor_unique_id, parse_raw_statements
+from warren.factories.base import BaseXapiStatementFactory
+
+
+def test_parse_raw_statements():
+    """Test the parsing of 2 simple statements, with the addition of a "date" column."""
+    statements = [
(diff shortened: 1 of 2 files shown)


src/backend/core/warren/tests/test_base_indicator.py
```diff
@@ -0,0 +1,94 @@
...
+            ]
+        ).dict(),
+    ]
+
+    statements = parse_raw_statements(statements)
+    statements = add_actor_unique_id(statements)
+    assert "actor.uuid" in statements.columns
+    assert statements["actor.uuid"].notna
+    # Check that 2 identical actors have the same UUID
+    uuids_john = statements[statements["actor.account.name"] == "John"]["actor.uuid"]
```
jmaupetit, line 93: I would ensure first that we have two matches.

```suggestion
    uuids_john = statements[statements["actor.account.name"] == "John"]["actor.uuid"]
    assert len(uuids_john) == 2
```

jmaupetit (commented):
:sunglasses: 

I only have a minor remark!

jmaupetit (approved):

PR #38, opened 2023-07-18
✅(backend) add unique views indicator test

cyrillay:
## Purpose

The recently merged indicators are lacking some unit tests. This PR adds one of them, and allowed to discover a potential issue in our implementation, see the comment on the PR.

## Proposal

Add a test for the video views endpoint, with the parameter `unique=True`.

### DIFF
--- a/src/api/plugins/video/tests/test_api.py
+++ b/src/api/plugins/video/tests/test_api.py
@@ -66,37 +66,38 @@ async def test_views_valid_video_id_path_but_no_matching_video(
 @pytest.mark.anyio
 async def test_views_backend_query(http_client: AsyncClient, httpx_mock: HTTPXMock):
     """Test the video views endpoint backend query results."""
-    # Build 3 video statement fixtures
-    view_1_date = "2020-01-01T00:00:00"
-    view_1_time = 100
-    view_2_date = "2020-01-01T00:00:30"
-    view_2_time = 200
-    view_3_date = "2020-01-02T00:00:00"
-    view_3_time = 300
+    # Define 3 video views fixtures
     video_uuid = "uuid://ba4252ce-d042-43b0-92e8-f033f45612ee"
+    video_views_fixtures = [
+        {"timestamp": "2020-01-01T00:00:00.000Z", "time": 100},
+        {"timestamp": "2020-01-01T00:00:30.000Z", "time": 200},
+        {"timestamp": "2020-01-02T00:00:00.000Z", "time": 300},
+    ]
+
+    # Build video statements from fixtures
     video_statements = [
         VideoPlayedFactory.build(
             [
                 {"object": {"id": video_uuid, "objectType": "Activity"}},
                 {"verb": {"id": PlayedVerb().id}},
-                {"result": {"extensions": {RESULT_EXTENSION_TIME: video_time}}},
-                {"timestamp": event_timestamp},
+                {"result": {"extensions": {RESULT_EXTENSION_TIME: view_data["time"]}}},
+                {"timestamp": view_data["timestamp"]},
             ]
         )
-        for video_time, event_timestamp in zip(
-            [view_1_time, view_2_time, view_3_time],
-            [view_1_date, view_2_date, view_3_date],
-        )
+        for view_data in video_views_fixtures
     ]
 
-    statements = [json.loads(statement.json()) for statement in video_statements]
+    # Convert each video statement to a JSON object
+    video_statements_json = [
+        json.loads(statement.json()) for statement in video_statements
+    ]
 
     # Mock the LRS call so that it returns the fixture statements
     lrs_client.base_url = "http://fake-lrs.com"
     httpx_mock.add_response(
         url=re.compile(r"^http://fake-lrs\.com/xAPI/statements\?.*$"),
         method="GET",
-        json={"statements": statements},
+        json={"statements": video_statements_json},
         status_code=200,
     )
 
@@ -112,8 +113,10 @@ async def test_views_backend_query(http_client: AsyncClient, httpx_mock: HTTPXMo
 
     assert response.status_code == 200
 
+    # Parse the response to obtain video views count_by_date
     video_views = (Response[VideoViews]).parse_obj(response.json()).content
 
+    # Counting all views is expected
     expected_video_views = {
         "total_views": 3,
         "count_by_date": [
--- a/src/api/plugins/video/tests/test_api.py
+++ b/src/api/plugins/video/tests/test_api.py
@@ -123,3 +123,75 @@ async def test_views_backend_query(http_client: AsyncClient, httpx_mock: HTTPXMo
     }
 
     assert video_views == expected_video_views
+
+
+@pytest.mark.anyio
+async def test_unique_views_backend_query(
+    http_client: AsyncClient, httpx_mock: HTTPXMock
+):
+    """Test the video views endpoint, with parameter unique=True."""
+    # Define 3 video views fixtures
+    video_uuid = "uuid://ba4252ce-d042-43b0-92e8-f033f45612ee"
+    video_views_fixtures = [
+        {"timestamp": "2020-01-01T00:00:00.000Z", "time": 100},
+        {"timestamp": "2020-01-01T00:00:30.000Z", "time": 200},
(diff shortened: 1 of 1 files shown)


src/backend/plugins/video/tests/test_api.py
```diff
@@ -123,3 +123,72 @@ async def test_views_backend_query(http_client: AsyncClient, httpx_mock: HTTPXMo
...
+    assert response.status_code == 200
+
+    video_views = (Response[VideoViews]).parse_obj(response.json()).content
+
+    expected_video_views = {
+        "total_views": 1,
+        "count_by_date": [
+            {"date": "2020-01-01", "count": 1},
+        ],
+    }
```
cyrillay, line 192: Currently, with `?unique=true`, if an actor has a view on the 2020-01-01 and on the 2020-01-03 (same actor), the last view is not counted because it was from the same actor.

Is this the behavior we want ?

Depending on the semantic of the indicator, we might want to remove duplicates only by day, so that here it would return 

`{‘total_views’=2, 'count_by_date': [{'date': '2020-01-01', 'count': 1}, {'date': '2020-01-02', 'count': 1}]}`
This means that uniqueness is only counted by a day basis. Hence, in the total, we could actually have views from the same actor.

WDYT ? If the current implementation is not right and we want the latter, I suggest opening an issue to fix it later.

Leobouloc: Perhaps a parameter `?first-viewing=true` could represent the current behavior ? Maybe the behavior you suggest could be `unique-actor` ?

cyrillay: I think it’s a bit weird to understand the `?first-viewing` from a user perspective… 

In my opinion, best solution would be to split the indicator to have : 
- one route that gives the `total_views` only
- one route that give the `count_by_date` only, and on this one, we count the unicity on a day basis, so that in the above example it would return `count_by_date': [{'date': '2020-01-01', 'count': 1}, {'date': '2020-01-02', 'count': 1}]}`

In both routes, `?unique` and `?complete` can be passed as parameters. WDYT ?

Leobouloc: I agree with having two routes (`views_by_date` would be a bit more consistent naming IMO). 

Perhaps `?unique-actor` would be more explicit ?

What is the purpose of `?complete` ?

cyrillay: Agree with you for the first option (two routes + name it `views_by_date`).

For the second one, `?unique-actor` is fine for me but I think we had discussed it before and consensus was to keep only `?unique` (not 100% sure though)

`?complete` is for counting completed views only, instead of normal views which can happen even if the actor didn't finish the video (xAPI verb `completed` instead of `played`)

jmaupetit: The unique video views indicator answers the following questions:

- how many different persons viewed my video?
- what about the dynamic of their first view?

So we want to keep it as it was implemented in the first attempt :wink:

cyrillay: So, let's keep it like this ? Got it 👍 

So, what about the rest of the PR ? Do you approve it ? 😉

lebrunthibault (approved):

src/api/plugins/video/tests/test_api.py
```diff
@@ -123,3 +123,72 @@ async def test_views_backend_query(http_client: AsyncClient, httpx_mock: HTTPXMo
...
     assert video_views == expected_video_views
+
+
+@pytest.mark.anyio
+async def test_unique_views_backend_query(
+    http_client: AsyncClient, httpx_mock: HTTPXMock
+):
+    """Test the video views endpoint, with parameter unique=True."""
+    # Build 3 video statement fixtures of the same actor
+    view_1_date = "2020-01-01T00:00:00"
```
lebaudantoine, line 134: Let's have one word per concept, either `date` or `timestamp`. I would go for the second one.

lebaudantoine, line 134: nit: to make it a real timestamp, should it have a timezone/ISO8601 string?
```suggestion
    view_1_date = "2020-01-01T00:00:00.000Z"
```

src/api/plugins/video/tests/test_api.py
```diff
@@ -123,3 +123,72 @@ async def test_views_backend_query(http_client: AsyncClient, httpx_mock: HTTPXMo
...
+                {"timestamp": event_timestamp},
+            ]
+        )
+        for video_time, event_timestamp in zip(
+            [view_1_time, view_2_time, view_3_time],
+            [view_1_date, view_2_date, view_3_date],
+        )
+    ]
+
+    statements = [json.loads(statement.json()) for statement in video_statements]
```
lebaudantoine, line 162: nit: misleading naming between `video_statements` and `statements`

src/api/plugins/video/tests/test_api.py
```diff
@@ -123,3 +123,72 @@ async def test_views_backend_query(http_client: AsyncClient, httpx_mock: HTTPXMo
...
+                        "account": {"name": "John", "homePage": "http://fun-mooc.fr"},
+                    }
+                },
+                {"object": {"id": video_uuid, "objectType": "Activity"}},
+                {"verb": {"id": PlayedVerb().id}},
+                {"result": {"extensions": {RESULT_EXTENSION_TIME: video_time}}},
+                {"timestamp": event_timestamp},
+            ]
+        )
+        for video_time, event_timestamp in zip(
```
lebaudantoine, line 156: Same here, these vars are related to a view?  `view_time` and `view_timestamp`

lebaudantoine (commented):
LGTM

src/api/plugins/video/tests/test_api.py
```diff
@@ -123,3 +123,72 @@ async def test_views_backend_query(http_client: AsyncClient, httpx_mock: HTTPXMo
...
+            [view_1_time, view_2_time, view_3_time],
+            [view_1_date, view_2_date, view_3_date],
+        )
+    ]
```
lebaudantoine, line 160:
```suggestion
      video_uuid = "uuid://ba4252ce-d042-43b0-92e8-f033f45612ee"
      video_views_fixtures = [
        {
            "date": "2020-01-01T00:00:00",
            "time": 100,
        },
        {
            "date": "2020-01-01T00:00:30",
            "time": 200,
        },
        {
            "date": "2020-01-02T00:00:00",
            "time": 300,
        },
    ]

    video_statements = [
        {
            "object": {"id": VIDEO_UUID, "objectType": "Activity"},
            "verb": {"id": "http://adlnet.gov/expapi/verbs/played"},
            "result": {"extensions": {RESULT_EXTENSION_TIME: view_data["time"]}},
            "timestamp": view_data["date"],
        }
        for view_data in video_views_fixtures
    ]
```

lebaudantoine, line 160: WDYT? 

```suggestion
    video_uuid = "uuid://ba4252ce-d042-43b0-92e8-f033f45612ee"
    video_views_fixtures = [
        {"date": "2020-01-01T00:00:00", "time": 100},
        {"date": "2020-01-01T00:00:30", "time": 200},
        {"date": "2020-01-02T00:00:00", "time": 300},
    ]
    
    # Build video statements from fixtures
    video_statements = [
        VideoPlayedFactory.build(
            [
                {
                    "actor": {
                        "objectType": "Agent",
                        "account": {"name": "John", "homePage": "http://fun-mooc.fr"},
                    }
                },
                {"object": {"id": video_uuid, "objectType": "Activity"}},
                {"verb": {"id": PlayedVerb().id}},
                {"result": {"extensions": {RESULT_EXTENSION_TIME: view_data["time"]}}},
                {"timestamp": view_data["date"]},
            ]
        )
        for view_data in video_views_fixtures
    ]
```

lebaudantoine (commented):
You could align `test_views_backend_query` code with the following approach to declare fixtures. 🌟

src/api/plugins/video/tests/test_api.py
jmaupetit: :100: :+1:

jmaupetit (approved):
LGTM

PR #47, opened 2023-07-25
🐛(api) return a count of 0 if there were no events

lebaudantoine:
## Purpose

The API should return a count of 0 if there were no events on a given date.

## Proposal

This fix is inspired by @jmaupetit previous works on the API, you could find [here](https://github.com/openfun/warren/blob/e1b60d5d0f4a35ce482872d2d3c62ba35deed49a/src/backend/plugins/video/warren_video/api.py).

During the indicator computation, it is essential to initialize the DailyCounts with a count of zero for each date within the expected date range. Subsequently, the counts should be updated with the appropriate values. This approach guarantees that each date will have a returned value, and the dates will be sorted accordingly.

- [x] Fix DailyVideoViews indicator
- [X] Fix DailyCompletedVideoViews indicator
- [x] Fix DailyVideoDownloads indicator.
- [x] Update indicators' tests.

### DIFF
--- a/src/api/core/warren/models.py
+++ b/src/api/core/warren/models.py
@@ -1,10 +1,14 @@
 """Warren's core models."""
 from enum import Enum
+from functools import reduce
+from itertools import groupby
 from typing import Any, Dict, Generic, List, TypeVar
 
+import arrow
 from pydantic.main import BaseModel
 
 from warren.fields import Date
+from warren.filters import DatetimeRange
 
 
 class StatusEnum(str, Enum):
@@ -39,9 +43,41 @@ class DailyCount(BaseModel):
     date: Date
     count: int = 0
 
+    def __add__(self, other):
+        """Add counters for two DailyCount instances with the same date."""
+        if self.date != other.date:
+            raise Exception("Cannot add two DailyCount instances with different dates")
+        return DailyCount(date=self.date, count=self.count + other.count)
+
 
 class DailyCounts(BaseModel):
     """Base model to represent daily counts summary."""
 
     total: int = 0
     counts: List[DailyCount] = []
+
+    @classmethod
+    def from_range(cls, date_range: DatetimeRange):
+        """Initialize DailyCounts from a date range."""
+        return cls(
+            counts=[
+                DailyCount(date=d)
+                for d in arrow.Arrow.range("day", date_range.since, date_range.until)
+            ]
+        )
+
+    def merge_counts(self, counts: List[DailyCount]):
+        """Merge DailyCount objects by date and aggregate counts.
+
+        This method allows us to combine a list of DailyCount objects into
+        the existing 'counts' list, aggregating the counts for each date
+        and ensuring that 'counts' contains only unique DailyCount objects
+        with one count per day.
+        """
+        self.counts += counts
+        self.counts.sort(key=lambda x: x.date)
+        self.counts = [
+            reduce(lambda x, y: x + y, v)
+            for k, v in groupby(self.counts, lambda dc: dc.date)
+        ]
+        self.total = sum(dc.count for dc in self.counts)
--- a/src/api/plugins/video/warren_video/indicators.py
+++ b/src/api/plugins/video/warren_video/indicators.py
@@ -10,7 +10,7 @@ from ralph.models.xapi.concepts.verbs.tincan_vocabulary import DownloadedVerb
 from ralph.models.xapi.concepts.verbs.video import PlayedVerb
 from warren.base_indicator import BaseIndicator, pre_process_statements
 from warren.filters import DatetimeRange
-from warren.models import XAPI_STATEMENT, DailyCounts
+from warren.models import XAPI_STATEMENT, DailyCount, DailyCounts
 from warren_video.conf import settings as video_plugin_settings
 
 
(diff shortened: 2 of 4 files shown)


src/api/plugins/video/warren_video/indicators.py
```diff
@@ -174,17 +175,24 @@ def compute(self) -> VideoViews:
...
-            .count()
-            .reset_index()
-            .rename(columns={"id": "count"})
-            .loc[:, ["date", "count"]]
-        )
+        # Calculate the count of events per date and rename the Series to "count"
+        count_by_date_with_views = flattened.groupby("date").size().rename("count")
+
+        # Generate a pd.Series of the expected date range.
+        expected_range_of_dates = pd.date_range(
```
lebaudantoine, line 182: `expected_date_range`?

jmaupetit: yup

jmaupetit (commented):
Looks great! 

A PR description would be lovely :wink:

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -174,17 +175,24 @@ def compute(self) -> VideoViews:
...
+            start=self.date_range.since.isoformat(),
+            end=self.date_range.until.isoformat(), # todo - discuss whether the end of range should be max to today's date
```
lebaudantoine, line 184: check whether last date is included

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -174,17 +175,24 @@ def compute(self) -> VideoViews:
...
+        # that missing dates will have a count of 0. The reset_index() method converts
+        # the Series to a DataFrame with the dates as a regular column.
+        count_by_date_with_nulls = count_by_date_with_views.reindex(
+            expected_range_of_dates, fill_value=0 # todo - discuss extracting this hardcoded value to a meaningful constant
+        ).reset_index()
 
         # Calculate the total number of events
         indicator.total_views = len(flattened.index)
-        indicator.count_by_date = count_by_date.to_dict("records")
+        indicator.count_by_date = count_by_date_with_nulls.to_dict("records")
```
lebaudantoine, line 196: when no raw statements, the count_by_date would be returned empty. Check @jmaupetit previous source code

lebaudantoine: https://github.com/openfun/warren/blob/e1b60d5d0f4a35ce482872d2d3c62ba35deed49a/src/backend/plugins/video/warren_video/api.py

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -172,18 +180,17 @@ def compute(self) -> DailyCounts:
...
+        # Calculate the count of events per date
+        count_by_date = flattened.groupby("date").size().items()
 
         # Calculate the total number of events
         indicator.total_count = len(flattened.index)
-        indicator.count_by_date = count_by_date.to_dict("records")
+
+        # Update the counts with the number of events per date
+        wip = {d.date: d.count for d in indicator.count_by_date}
+        wip.update(count_by_date)
```
lebaudantoine, line 191: update this naming

lebaudantoine:
Feedback: The `DailyCounts` model poses challenges in manipulation, impacting the developer experience negatively due to:

- Difficulty in updating date counts.
- Lack of default initialization from a date range or list of dates.
- Absence of a mechanism to prevent duplicate counts for the same date.
- Absence of mechanism to make sure date counts are sorted in the right order.

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -69,7 +69,9 @@ def compute(self) -> DailyCounts:
         Fetches the statements from the LRS, filters and aggregates them to return the
         number of video views per day.
         """
-        indicator = DailyCounts()
+        # Initialize daily counts within the specified date range,
+        # with counts initialized to zero
+        indicator = DailyCounts.from_range(self.date_range)
```
jmaupetit, line 74: This is way more elegant :clap:

src/api/core/warren/models.py
```diff
@@ -39,9 +43,47 @@ class DailyCount(BaseModel):
...
+    def from_range(cls, date_range: DatetimeRange):
+        """Initialize DailyCounts from a date range."""
+        return cls(
+            count_by_date=[
+                DailyCount(date=d)
+                for d in arrow.Arrow.range("day", date_range.since, date_range.until)
+            ]
+        )
+
+    def merge(self, counts: List[DailyCount]):
```
jmaupetit, line 75: It wouldn't change that much, but I find it more intuitive to merge two objects instead of an object and a specific field of another instance of the same object type. WDYT?

```suggestion
    def merge(self, other: DailyCounts):
```

jmaupetit (approved):
Approved to let you pursue your work while leaving for few weeks, but pay attention to my comments :wink:

src/api/core/warren/models.py
lebaudantoine: Initially I tried the following, however while using it in the `compute` method, I figured out I had to instantiate a `DailyCounts` object just to pass a List `DailyCount`. Having two methods when we need it could be great? I could rename it to `merge_counts`, WDYT? 

```
    def merge(self, other: DailyCounts):
        return self.merge_counts(other.counts)

    def merge_counts(self, other: List[DailyCount]):
    ...
    
```

jmaupetit: I see. That would be an acceptable solution!

PR #49, opened 2023-07-26
♻️(api) factorize video's indicators

lebaudantoine:
## Purpose

Reduce code's duplication between video's indicators.

## Proposal

Abstract the views and downloads indicators into a more modular general indicator. 

- [x] Offer a reusable indicator class.
- [x] Provide additional feature to the indicator using mixin.
- [x] Update documentation
- [x] Update API routes
- [x] Update tests

### DIFF
--- a/src/api/core/warren/base_indicator.py
+++ b/src/api/core/warren/base_indicator.py
@@ -7,66 +7,88 @@ import pandas as pd
 from ralph.backends.http.async_lrs import LRSQuery
 
 from warren.conf import settings
+from warren.models import XAPI_STATEMENT
 
 
-def parse_raw_statements(raw_statements) -> pd.DataFrame:
-    """Parse LRS statements, explode the columns, and add a `date` column."""
-    flattened = pd.json_normalize(raw_statements)
-    # Disable chained assignment warning to make the transformation inplace
-    pd.options.mode.chained_assignment = None
-    # Transform timestamp column into a date with day (YYYY-MM-DD)
-    flattened.loc[:, "date"] = pd.to_datetime(flattened.loc[:, "timestamp"]).dt.date
-    return flattened
+class PreprocessMixin(object):
+    """Mixin class for preprocessing xAPI statements.
 
-
-def add_actor_uid(statements: pd.DataFrame) -> pd.DataFrame:
-    """Add a `actor.uid` column that uniquely identifies the agent.
-
-    Depending on the xAPI statements, the actor can be identified in 4 different ways :
-    https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#details-4. This
-    function handles the 4 cases and creates a `uid` column that can be used later
-    without worrying about the 4 IFIs.
+    This class provides methods to parse LRS statements, add a 'date' column,
+    and add an 'actor.uid' column that uniquely identifies the agent.
+    It also includes a method to preprocess the statements and apply all
+    the necessary transformations all at once.
     """
-    xapi_actor_identifier_columns = settings.XAPI_ACTOR_IDENTIFIER_PATHS.intersection(
-        set(statements.columns)
-    )
 
-    if len(xapi_actor_identifier_columns) == 0:
-        raise ValueError(
-            "There is no way of identifying the agent in submitted statements."
+    raw_statements: List[XAPI_STATEMENT] = None
+    statements: pd.DataFrame = None
+
+    def parse_raw_statements(self) -> None:
+        """Parse LRS statements, explode the columns."""
+        self.statements = pd.json_normalize(self.raw_statements)
+
+    def add_date_column(self) -> None:
+        """Add a 'date' column computed from statement's timestamp."""
+        # Disable chained assignment warning to make the transformation inplace
+        with pd.option_context("mode.chained_assignment", None):
+            # Transform 'timestamp' column into a date with the format (YYYY-MM-DD)
+            self.statements["date"] = pd.to_datetime(
+                self.statements["timestamp"]
+            ).dt.date
+
+    def add_actor_uid_column(self) -> None:
+        """Add a 'actor.uid' column that uniquely identifies the agent.
+
+        Depending on the xAPI statements, the actor can be identified in 4 ways:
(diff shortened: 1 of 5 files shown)


src/api/core/warren/base_indicator.py
```diff
@@ -64,9 +82,13 @@ def get_lrs_query(self) -> LRSQuery:
...
 
     @abstractmethod
-    def fetch_statements(self) -> List:
+    def fetch_statements(self):
         """Executes the LRS query to obtain statements required for the indicator."""
 
+    @abstractmethod
+    def filter_statements(self):
+        """Filters the statements required for the indicator."""
+
```
lebaudantoine, line 91: Should it be an `@abstractmethod` from `BaseIndicator` or added/provided with inheritance of a `FilterMixin`? I am open to discuss it.

src/api/plugins/video/warren_video/api.py
```diff
@@ -28,26 +30,39 @@ async def views(
...
+    query_parameters = VideoQueryParameters(
+        video_id=video_id,
+        verb_id=verb_id,
+        date_range=DatetimeRange(since=filters.since, until=filters.until),
+    )
+    indicator = DailyVideoEventIndicator(
+        client=lrs_client,
+        query_parameters=query_parameters,
+        remove_duplicate_actors=unique,
+        statements_filter=statements_filter,
```
lebaudantoine, line 56: Should we group those arguments into a single structure, such as `filter_parameters`, to show that they're all used in the same method?

src/api/plugins/video/warren_video/api.py
```diff
@@ -28,26 +30,39 @@ async def views(
...
-        indicator = DailyVideoViews(**indicator_kwargs)
+        # todo - extract this filter definition?
+        def filter_statements_on_duration(row):
+            return (
+                row[f"result.extensions.{RESULT_EXTENSION_TIME}"]
+                >= video_plugin_settings.VIEWS_COUNT_TIME_THRESHOLD
+            )
+
+        verb_id = PlayedVerb().id
+        statements_filter = filter_statements_on_duration
```
lebaudantoine, line 45: It feels like the route deeply change its behavior based on a `if/else`

src/api/plugins/video/warren_video/api.py
```diff
@@ -28,26 +30,39 @@ async def views(
...
+        query_parameters=query_parameters,
+        remove_duplicate_actors=unique,
+        statements_filter=statements_filter,
+    )
     try:
-        response = Response[DailyCounts](
-            status=StatusEnum.SUCCESS, content=indicator.compute()
-        )
+        result = indicator.compute()
+        response = Response[DailyCounts](status=StatusEnum.SUCCESS, content=result)
```
lebaudantoine, line 60: could be condensed back? WDYT ?

src/api/plugins/video/warren_video/api.py
```diff
@@ -58,19 +73,22 @@ async def downloads(
...
+        client=lrs_client,
+        query_parameters=query_parameters,
+        remove_duplicate_actors=unique,
     )
     try:
-        response = Response[DailyCounts](
-            status=StatusEnum.SUCCESS, content=indicator.compute()
-        )
+        result = indicator.compute()
+        response = Response[DailyCounts](status=StatusEnum.SUCCESS, content=result)
```
lebaudantoine, line 88: ditto.

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -1,269 +1,103 @@
...
-from warren.models import XAPI_STATEMENT, DailyCounts
-from warren_video.conf import settings as video_plugin_settings
+from warren.base_indicator import BaseIndicator, PreprocessMixin
+from warren.models import DailyCounts
+from warren_video.models import VideoQueryParameters
 
 
-class DailyVideoViews(BaseIndicator):
-    """Daily Video Views indicator.
+class DailyVideoEventIndicator(BaseIndicator, PreprocessMixin):
```
lebaudantoine, line 11: is `Event` the right naming? Anything closer to the xAPI semantic?

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -1,269 +1,103 @@
...
-
-        # Calculate the total number of events
-        indicator.total_count = len(filtered_view_duration.index)
-        indicator.count_by_date = count_by_date.to_dict("records")
-
-        return indicator
+        if self.statements_filter:
+            self.statements = self.statements[
+                self.statements.apply(self.statements_filter, axis=1)
+            ]
```
lebaudantoine, line 70: Does this provide sufficient modularity? 
Should we have another indicator class defined, that calls the parent method and surcharges it to handle some custom filters?

lebaudantoine (commented):
Let's discuss some parts of this refactoring ! I would appreciate some feedbacks on the `filter_statements` parts.

src/api/core/warren/base_indicator.py
jmaupetit: An abstract method from the `BaseIndicator` seems fine to me.

src/api/plugins/video/warren_video/models.py
```diff
@@ -1 +1,12 @@
...
+from pydantic.main import BaseModel
+from warren.filters import DatetimeRange
+
+
+class VideoQueryParameters(BaseModel):
+    """Base model to represent an LRS query for fetching video statements."""
+
+    video_id: str
+    verb_id: str
+    date_range: DatetimeRange
```
jmaupetit, line 12: If not used in the API request signature, it seems fragile to me.

jmaupetit (commented):
Interesting pattern! Don't miss the point: API views should be as dumb as possible, an indicator should be self-sufficient and explicit; it can be used in various contexts, outside from the API, such as cron task, etc.

jmaupetit (commented):
It looks way better to me. Cannot wait to see all your current improvements merged!

I'll let @quitterie-lcs and @wilbrdt validate this work while I'm away. Thanks!

wilbrdt (approved):
That's nice and clean! I like it :100: 
Just nit-picking the docstrings imperative style but good to go! :tada:

PR #60, opened 2023-08-10
✅(backend) add persistence layer

cyrillay:
Add pgsql service to save computed indicators values

## Purpose

Some indicators might require heavy computations, and the might be accessed multiple times for the same value, which currently triggers a recomputation each time. For some indicators, we would like to offer a way to persist them in a database, so that the value can be retrieved if it was already computed. Also, this will prepare the work for scheduled computing of the indicators : for example an indicator can be computed everyday at 1am and persisted in this DB.

## Proposal

- Add a postgresql service where we can store the indicator's values. 
- Add mechanism to save and retrieve values


src/api/plugins/video/warren_video/api.py
```diff
@@ -40,6 +40,7 @@ async def views(
         response = Response[DailyCounts](
             status=StatusEnum.SUCCESS, content=await indicator.compute()
         )
+        await indicator.persist()
```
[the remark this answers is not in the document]
lebaudantoine: Great approach! !! After a quick review, how do you retrieve these stored values to prevent recomputing the indicator?

lebaudantoine (commented):
Great approach!!

docker-compose.yml
```diff
@@ -6,6 +6,17 @@ services:
...
+  postgresql-indicators:
+    image: postgres:12
+    ports:
+      - "${WARREN_POSTGRES_INDICATORS_PORT:-5432}:${WARREN_POSTGRES_INDICATORS_PORT:-5432}"
+    volumes:
+      - ./pgdata:/var/lib/postgresql/data
+    env_file:
+      - .env
+    environment:
+      POSTGRES_PASSWORD: ${WARREN_POSTGRES_INDICATORS_PASSWORD}
```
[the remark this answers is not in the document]
cyrillay: Agreed, thanks !

src/api/core/warren/conf.py
```diff
@@ -40,6 +40,11 @@ class Settings(BaseSettings):
         "actor.openid",
     }
 
+    # Persistence configuration
+    POSTGRES_INDICATORS_USER: str = "postgres"
+    POSTGRES_INDICATORS_PORT: int = 5432
```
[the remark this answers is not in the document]
cyrillay: So in the `.env.dist` you mean ? Yes for sure, you are right !

src/api/core/warren/database_manager.py
```diff
@@ -0,0 +1,106 @@
...
+                self.connection = psycopg2.connect(
+                    user=settings.POSTGRES_INDICATORS_USER,
+                    password=settings.POSTGRES_INDICATORS_PASSWORD,
+                    host="postgresql-indicators",
+                    port=settings.POSTGRES_INDICATORS_PORT,
+                    database="postgres"
+                )
+                self.cursor = self.connection.cursor()
+            except (Exception, psycopg2.Error) as error:
+                print("Error while connecting to PostgreSQL:", error)
```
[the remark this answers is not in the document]
cyrillay: Agreed as well !

src/api/core/warren/database_manager.py
```diff
@@ -0,0 +1,106 @@
...
+            self.connection.close()
+            self.connection = None
+            self.cursor = None
+
+    def execute_query(self, query, params=None):
+        try:
+            self.cursor.execute(query, params)
+            self.connection.commit()
+        except (Exception, psycopg2.Error) as error:
+            print("Error executing query:", error)
```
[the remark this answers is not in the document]
cyrillay: Definitely ! Good catch

src/api/core/warren/database_manager.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def connect(self):
+        if not self.connection:
+            try:
+                self.connection = psycopg2.connect(
+                    user=settings.POSTGRES_INDICATORS_USER,
+                    password=settings.POSTGRES_INDICATORS_PASSWORD,
+                    host="postgresql-indicators",
+                    port=settings.POSTGRES_INDICATORS_PORT,
+                    database="postgres"
```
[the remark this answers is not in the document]
cyrillay: I hesitated at first in order to avoid making too much config, but in the end I agree with you ;)

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,6 +82,33 @@ def filter_statements(self) -> None:
...
+                "(%s, %s, %s) ON CONFLICT DO NOTHING",
+                (self.date_range.since.isoformat(), self.date_range.until.isoformat(), self.computed_indicator.total)
+            )
+            logger.debug(f"Successfully saved ${cursor.rowcount} records in date_range_views_count.")
+
+            # Insert daily counts
+            for count in self.computed_indicator.counts:
+                cursor.execute("INSERT INTO daily_views_count"
+                               "(date, count)"
+                               "VALUES (%s, %s) ON CONFLICT DO NOTHING", (count.date, count.count))
```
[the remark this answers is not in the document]
cyrillay: That's a good point, but I'm thinking of deleting all the CRUD methods I created, and instead directly use the cursor and db connection all the time. I think it doesn't add value to have a wrapper for those CRUD methods, and we lose the ability to fine-control when to commit (after the individual insertions of the for loop in the above case)
I'm also thinking, if such a wrapper had value, `psycopg2` would have added it 😂

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,6 +82,33 @@ def filter_statements(self) -> None:
...
+            logger.debug(f"Successfully saved ${cursor.rowcount} records in daily_views_count.")
```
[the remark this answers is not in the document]
cyrillay: I added it, you're right it was missing. But there is also the issue of conflicts between video views that are : 
- unique
- complete
- unique & complete

For now I didn't think that much about this part, and for now, I don't have an elegant solution in my mind to handle the cases with all model designs. I work on it and get back to you 👍

src/api/plugins/video/warren_video/api.py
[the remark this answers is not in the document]
cyrillay: I made it optional, but in a simple way for now, we will probably need more granularity later (persist some indicators only, and don't persist some others)

Regarding the retrieval, this is not implemented for now. Lots of things are missing in the current PR state, I wanted early feedback before being too deep in the implementation ;)

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,6 +82,33 @@ def filter_statements(self) -> None:
         if self.remove_duplicate_actors:
             self.statements.drop_duplicates(subset="actor.uid", inplace=True)
 
+    async def persist(self) -> None:
+        if not self.computed_indicator:
+            raise Exception("Indicator must be computed before being persisted.")
```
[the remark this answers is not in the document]
cyrillay: Yes, you are right it wouldn't be caught. As we had discussed on slack, I think it would be nice to do a refactor of the error handling (use HTTP codes/fast-api builtin exceptions for instance)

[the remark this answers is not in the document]
lebaudantoine: yes, but that's two different topics ! I would separate this in two PRs

src/api/plugins/video/warren_video/indicators.py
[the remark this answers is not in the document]
lebaudantoine: What about having an UID, like a string, based on : 

- Indicator id / name
- Indicators params
- Resource uid

src/api/plugins/video/warren_video/api.py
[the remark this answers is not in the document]
lebaudantoine: Actually, could it be better to call this method, and the one retrieving the persistent data, in the `compute` method. To keep the api file as dumb as possible, and make sure the indicator has been computed before persist it

src/api/plugins/video/warren_video/indicators.py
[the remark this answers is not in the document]
lebaudantoine: @wilbrdt I think there is a bit of software architecture and OOP there, what would you think of a dedicated Mixins class to handle any logic related to the persistence?

src/api/plugins/video/warren_video/indicators.py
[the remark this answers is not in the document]
lebaudantoine: However, I don't really like duplicated constant literals such `date_range_views_count`. It's error-prone. I would suggest having at least few constants to avoid any mistake while working with the db schema.

src/api/plugins/video/warren_video/api.py
```diff
@@ -40,6 +42,8 @@ async def views(
         response = Response[DailyCounts](
             status=StatusEnum.SUCCESS, content=await indicator.compute()
         )
+        if settings.WARREN_IS_PERSISTENCE_ENABLED:
+            await indicator.persist()
```
wilbrdt, line 46: It would greatly slow down the API response time if done here. Could we add it as a [background task](https://fastapi.tiangolo.com/tutorial/background-tasks/)?

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,6 +84,33 @@ def filter_statements(self) -> None:
...
+
+            # Insert daily counts
+            for count in self.computed_indicator.counts:
+                cursor.execute("INSERT INTO video_daily_views"
+                               "(video_id, date, count)"
+                               "VALUES (%s, %s, %s) ON CONFLICT DO NOTHING", (self.video_id.replace("uuid://", ""), count.date, count.count))
+            logger.debug(f"Successfully saved ${cursor.rowcount} records in daily_views_count.")
+
+            db.commit()
+        db_manager.disconnect()
```
wilbrdt, line 112: As it won't be the only indicator that we persists in a database, every indicator would have to repeat these steps. We could probably add a generic helper that takes care of connecting to the database of choice, create/insert, then disconnect. WDYT?

wilbrdt (commented):
Promising! :rocket:

src/api/plugins/video/warren_video/indicators.py
lebaudantoine: I tagged you in a comment, about creating a Mixin class, in charge of it. Let's discuss it

src/api/plugins/video/warren_video/indicators.py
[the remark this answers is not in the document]
wilbrdt: Yeah I agree! How about ORM style, having a table for each indicator: table `DailyViews`, table `DailyCompletedViews` etc
With this, in each table, we could store:
- an id, based on indicators params
- the computed value(s). Could be a simple int, or a JSON fields
WDYT?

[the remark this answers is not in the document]
wilbrdt: The problem with what I proposed: we still need to store the `date_range` in another column so we can build upon what we computed in the past. What about storing the whole indicator class as JSON fields. We would have as many columns as there are attributes for the indicator (with useless ones, e.g. `client`) but we could query over `date_range` etc

src/api/plugins/video/warren_video/api.py
cyrillay: Yes, that's a good idea. The only thing that I'm worried about is the monitoring : if something fails in the background task, will we see it ?
I quickly tested and this call seemed to take <10% of the overall time spent for the computation and response, but it should be benchmarked more thoroughly. I test the background task to see if it works fine 👍

src/api/plugins/video/warren_video/indicators.py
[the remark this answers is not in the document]
cyrillay: Thanks @wilbrdt, this is exactly the kind of question I was trying to answer !

The thing is that currently, the `DailyCounts` model contains a field `counts`, which is a `List[DailyCount]`, basically a list of date->count tuples. And, this list can be big : with the current implementation, for a 3 months date range, this gives a list of 3*31 tuples, and all of this would be stored under a common field in the `DailyViews` table, which is not ideal for a single field.

Although we won't hit any postgres limit for the size of a field, this limits our ability to do incremental work on the indicator : if we have persisted a 3 months range, and the next query is on that same range plus one day, it's hard to modify the big field in the DB (and requires indicator-specific work)... What do you think ?

Thanks for your input though 👍

src/api/core/warren/conf.py
```diff
@@ -40,6 +40,14 @@ class Settings(BaseSettings):
         "actor.openid",
     }
 
+    # Persistence configuration
+    WARREN_IS_PERSISTENCE_ENABLED: str = False
```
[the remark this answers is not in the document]
cyrillay: Good catch ! Thanks

src/api/plugins/video/warren_video/indicators.py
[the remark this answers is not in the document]
cyrillay: Yes, we can set the column names in constants. For now I don't do it, as the schema and column/table names are still evolving, but I keep this as unresolved so that we know to do it.

src/api/plugins/video/warren_video/api.py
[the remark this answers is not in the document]
cyrillay: I agree with you, the api file must be as simple as possible. I modified it in a fixup commit, let me know what you think

src/api/core/warren/exceptions.py
```diff
@@ -0,0 +1,4 @@
+
+class UncomputedIndicatorError(Exception):
+    """Raised when an indicator is not computed before persisting."""
```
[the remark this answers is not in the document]
cyrillay: In the end, after a comment from @lebaudantoine, I modified the logic so that the `.persist()` is called inside the `.compute()`, so this exception doesn't have any reason to be thrown anymore, I delete the file

src/api/core/warren/database_manager.py
```diff
@@ -0,0 +1,58 @@
...
+
+import psycopg2
+
+from warren.conf import settings
+
+logger = logging.getLogger(__name__)
+
+class DatabaseManager(ABC):
+    @abstractmethod
+    def connect(self):
```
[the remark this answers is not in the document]
cyrillay: Yep, I haven't dealt with all the linter issues for now, I want to make sure everyone agrees on the current design first ;) I keep this comment opened to make sure I don't forget it.

src/api/plugins/video/warren_video/indicators.py
cyrillay: For now, those steps are shared for all the indicators that inherit from `BaseDailyEvent`, so all our current indicators. If we later define other types of indicators, I think it will be a good time to factorize the logic in generic helpers. WDYT ?

src/api/plugins/video/warren_video/api.py
cyrillay: Well, I tested it and it worked like a charm, logs are forwarded correctly. Plus, we now return a correct response even if an exception happens during the persistence step. Nice 👍

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,12 +90,72 @@ def filter_statements(self) -> None:
...
+            logger.info("Indicator persistence is disabled, skipping...")
+            return
+        logger.info("Indicator persistence is enabled, storing computed indicator...")
+        db_manager = PgsqlManager()
+        db_manager.connect()
+        db = db_manager.connection
+        with db.cursor() as cursor:
+            # Insert total of the range
+            cursor.execute(
+                "INSERT INTO indicator_date_range_events"
```
cyrillay, line 104: Open question : how/where to setup the schemas of the database that should be executed only after creation of the DB ? I currently use : 

```
CREATE TABLE indicator_daily_events (
	indicator_uuid VARCHAR(65),
	video_id UUID,
    date DATE,
    count INT,
    PRIMARY KEY(indicator_uuid, date)
);

CREATE TABLE indicator_date_range_events (
	indicator_uuid VARCHAR(65),
	video_id UUID,
    range_start DATE,
    range_end DATE,
    count INT,
    PRIMARY KEY (indicator_uuid)
);
```

.env.dist
```diff
@@ -27,3 +27,11 @@ RALPH_AUTH_FILE=/app/.ralph/auth.json
 RALPH_BACKENDS__DATABASE__ES__HOSTS=http://elasticsearch:9200
 RALPH_BACKENDS__DATABASE__ES__INDEX=statements
 RALPH_RUNSERVER_PORT=8200
+
+# Postgresql configuration (Indicators persistence)
+WARREN_IS_PERSISTENCE_ENABLED=True
+WARREN_POSTGRES_INDICATORS_HOST="postgresql-indicators"
+WARREN_POSTGRES_INDICATORS_PORT=5432
+WARREN_POSTGRES_INDICATORS_DATABASE="postgres"
```
lebaudantoine, line 35: Nit: How about using a more precise term like "indicators" or "persistency" to align better with the "lti" database nomenclature?

.env.dist
```diff
@@ -27,3 +27,11 @@ RALPH_AUTH_FILE=/app/.ralph/auth.json
...
+WARREN_POSTGRES_INDICATORS_USER="postgres"
+WARREN_POSTGRES_INDICATORS_PASSWORD=password
```
lebaudantoine, line 37: Nit: Would it be a bad practice to use the same username and password as the LTI one?

docker-compose.yml
```diff
@@ -6,6 +6,17 @@ services:
     env_file:
       - .env
 
+  postgresql-indicators:
```
lebaudantoine, line 9: Nit: Should we rename, the other `postgresql` to `postgresql-lti`, actually I am indecisive ..

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -44,6 +51,12 @@ def __init__(
         self.remove_duplicate_actors = remove_duplicate_actors
         self.video_id = video_id
         self.date_range = date_range
+        self.computed_indicator: DailyCounts | None = None
+        self.uuid = hashlib.sha256(
+                "-".join(
+                    col for col in [self.video_id, str(self.date_range), str(self.remove_duplicate_actors), self.__class__.__name__]
+                ).encode()
+            ).hexdigest()
```
lebaudantoine, line 59: Is there any reason that would justify computing it in the constructor?
I would extract these in a cached property, or at least in a private method.

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,12 +90,72 @@ def filter_statements(self) -> None:
...
+                    logger.info("Indicator not found in DB, computing...")
+                    return False
+
     async def compute(self) -> DailyCounts:
         """Fetch statements and computes the current indicator.
 
         Fetch the statements from the LRS, filter and aggregate them to return the
         number of video events per day.
         """
+        if self.fetch_persisted_indicator():
```
lebaudantoine, line 156: Great, we might miss a mechanism to force computing again the indicator, WDYT?

src/api/plugins/video/warren_video/indicators.py
lebaudantoine: Do we agree that the tables should initialize while bootstrapping the project?

.env.dist
```diff
@@ -27,3 +27,11 @@ RALPH_AUTH_FILE=/app/.ralph/auth.json
 RALPH_BACKENDS__DATABASE__ES__HOSTS=http://elasticsearch:9200
 RALPH_BACKENDS__DATABASE__ES__INDEX=statements
 RALPH_RUNSERVER_PORT=8200
+
+# Postgresql configuration (Indicators persistence)
+WARREN_IS_PERSISTENCE_ENABLED=True
+WARREN_POSTGRES_INDICATORS_HOST="postgresql-indicators"
```
lebaudantoine, line 33: Nit: is it really necessary to prefix these variables with the project names? It ends up in really long var names.

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,12 +90,72 @@ def filter_statements(self) -> None:
...
+        db_manager.disconnect()
+
+    def fetch_persisted_indicator(self):
+        """Check if the indicator was already persisted. If so, retrieve the value."""
+        if not settings.IS_PERSISTENCE_ENABLED:
+            logger.info("Persistence disabled, computing the indicator...")
+        else:
+            db_manager = PgsqlManager()
+            db_manager.connect()
+            with db_manager.cursor as cursor:
```
lebaudantoine, line 128: These lines are a bit duplicated with l.98-101. What about the Mixins?

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,12 +90,72 @@ def filter_statements(self) -> None:
...
+                               "VALUES (%s, %s, %s, %s) ON CONFLICT DO NOTHING", (self.uuid, self.video_id.replace("uuid://", ""), count.date, count.count))
+            logger.info(f"Successfully saved {len(self.computed_indicator.counts)} records in indicator_daily_events.")
+
+            db.commit()
+        db_manager.disconnect()
+
+    def fetch_persisted_indicator(self):
+        """Check if the indicator was already persisted. If so, retrieve the value."""
+        if not settings.IS_PERSISTENCE_ENABLED:
+            logger.info("Persistence disabled, computing the indicator...")
```
lebaudantoine, line 124: Return early to avoid long nested code?

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,12 +90,72 @@ def filter_statements(self) -> None:
...
+        else:
+            db_manager = PgsqlManager()
+            db_manager.connect()
+            with db_manager.cursor as cursor:
+                cursor.execute(
+                    "SELECT * FROM indicator_date_range_events WHERE indicator_uuid = %s",
+                    (self.uuid,)
+                )
+                stored_total = cursor.fetchone()
```
lebaudantoine, line 133: General comment, is it really worth it to make the total persistent, while it can be easily guess from daily counts? Let's have a discussion

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,12 +90,72 @@ def filter_statements(self) -> None:
...
+                stored_by_day = cursor.fetchall()
+                if stored_total and stored_by_day:
+                    logger.info("Found computed indicator in DB, retrieving...")
+                    self.computed_indicator = DailyCounts(total=stored_total['count'],
+                                                          counts=[DailyCount(date=d['date'], count=d['count']) for d in
+                                                                  stored_by_day])
+                    return True
+                else:
+                    logger.info("Indicator not found in DB, computing...")
+                    return False
```
lebaudantoine, line 148: in `persist` you disconnect from the db_manager, `db_manager.disconnect()`, it should be necessary there also?

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,12 +90,72 @@ def filter_statements(self) -> None:
...
+        db_manager = PgsqlManager()
+        db_manager.connect()
+        db = db_manager.connection
+        with db.cursor() as cursor:
+            # Insert total of the range
+            cursor.execute(
+                "INSERT INTO indicator_date_range_events"
+                "(indicator_uuid, video_id, range_start, range_end, count) VALUES"
+                "(%s, %s, %s, %s, %s) ON CONFLICT DO NOTHING",
+                (self.uuid, self.video_id.replace("uuid://", ""), self.date_range.since.isoformat(), self.date_range.until.isoformat(), self.computed_indicator.total)
```
lebaudantoine, line 107: I'm curious as to why we're not considering inserting only the `uuid` and the `count`?

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -77,12 +90,72 @@ def filter_statements(self) -> None:
...
+            )
+            logger.info(f"Successfully saved {cursor.rowcount} record in indicator_date_range_events.")
+
+            # Insert daily counts
+            for count in self.computed_indicator.counts:
+                cursor.execute("INSERT INTO indicator_daily_events"
+                               "(indicator_uuid, video_id, date, count)"
+                               "VALUES (%s, %s, %s, %s) ON CONFLICT DO NOTHING", (self.uuid, self.video_id.replace("uuid://", ""), count.date, count.count))
```
lebaudantoine, line 115: ditto.

src/api/plugins/video/warren_video/indicators.py
lebaudantoine: it can be discussed tho.

lebaudantoine (commented):
niceee 🎉

src/api/plugins/video/warren_video/indicators.py
cyrillay: I agree !

src/api/plugins/video/warren_video/indicators.py
cyrillay: Good catch ! I add it

src/api/plugins/video/warren_video/indicators.py
cyrillay: For this one, we could actually store `uuid` and `count` only, the only reason I store the other fields is for human readability, so that we can know while looking at the DB table what dates the rows are about.

src/api/plugins/video/warren_video/indicators.py
cyrillay: For this one, we need at least one more field such as the date to identify the row, as one indicator ID will have many values covering the date range (one for each date). The other fields are here for the same reason as above : for human readability, but this can be discussed

src/api/plugins/video/warren_video/indicators.py
cyrillay: Yes, good one 👍

src/api/plugins/video/warren_video/indicators.py
cyrillay: Well, it can be "guessed" but it still needs an additional SQL query with a groupby. It's not mandatory to have as we can still recompute it, but it doesn't hurt to have it IMO

src/api/plugins/video/warren_video/indicators.py
cyrillay: IMO, 3 lines of duplication isn't a big deal, but it can be discussed ! You suggest to create an object like `Postprocess`mixin and have the `.persist()` (or `.fetch_persisted()`) method there ? The only thing is that we need to use the indicator fields, so everything will need to be passed as parameters between the indicator and its mixin

.env.dist
cyrillay: I am following [this](https://github.com/openfun/warren/blob/81d3e42228be068ef8a0d3cad3e0c5d08f077592/src/api/core/warren/conf.py#L61) convention, but in the end I agree with you, all the conf variables in warren are about warren, only 4 about ralph, so we could ditch the `WARREN_` prefix

docker-compose.yml
cyrillay: I think we should : Now that we have more than one DB, each should be suffixed with the component its about.

.env.dist
cyrillay: Not a bad practices, but not a great one neither :p 

Currently, I don't see any configured for the `postgres-lti` app 🤔

docker-compose.yml
lebaudantoine: ok nice !

src/api/plugins/video/warren_video/indicators.py
lebaudantoine: Like a "force" option, it's quite common for cached data to have a proper way to clean/ignore the cached value.
Anyway, not essential, let's focus on what matters.

.env.dist
lebaudantoine: nit-picking comment, you decide !

src/api/plugins/video/warren_video/indicators.py
lebaudantoine: Or a simple python `sum`. With pydantic v2, the total would evolve in a property, and thus won't be set while instantiating the `DailyCounts`.

lebaudantoine: I would choose a minimalist table schema, less code !

src/api/plugins/video/warren_video/indicators.py
lebaudantoine: I buy the human readability 🤗

src/api/plugins/video/warren_video/indicators.py
lebaudantoine: alright ! Your call, you are the expert

.env.dist
cyrillay: You're right, I modify !

src/api/plugins/video/warren_video/indicators.py
lebaudantoine: I would make it more general, as a `DatabaseMixin`, that add database capabilities to any class, namely 

- connection
- disconnection
- CRUD operation

We could define an interface, that would define what are the major method of a cache tool, and thus decoupling this indicator from the actual PSQLManager, passing any cache tool should work. In the longer run, we would have only to update the mixin.

.env.dist
lebaudantoine: there are some actually

lebaudantoine: [l21](https://github.com/openfun/warren/blob/156069720d16d2d28902cd45d431b3957856285c/.env.dist#L21)
[l13](https://github.com/openfun/warren/blob/156069720d16d2d28902cd45d431b3957856285c/.env.dist#L12)

src/api/plugins/video/warren_video/indicators.py
cyrillay: I don't have any opinion on this one, so I make it a cached property

jmaupetit:
This work has been revamped in #79

PR #69, opened 2023-08-17
✨(frontend) Basic LTI rendering 

lebaudantoine:
## Purpose

Integrate the Web app using the Django LTI server for rendering.

## Proposal

This has been influenced by Marsha's approach, but please keep in mind that this implementation is not yet finalized. Our upcoming plans involve substantial code refactoring to incorporate app routing and refine the initialization logic.

- [X] Output frontend build in Django static files.
- [X] Inject the JS bundle and stylesheet in Django base template.
- [X] Render the SPA in a dedicated div.
- [X] Implement data transmission to the frontend through the DOM.

#### Dev environment

Running `make run` would start the development server, namely the Django app. Then, you could access the frontend through two paths:

- The dashboard navigating through `http://localhost:8090/dashboards/`
- The development interface refined by @jmaupetit navigating through `http://localhost:8090/develoment/`.

#### Anticipated Improvements and Features:
In the upcoming iterations, we aspire to set up the frontend app initialization process and content loading. Our app's capabilities, while limited, will allow users to:

- Select specific resources (Dashboard or plot) through deep linking.
- Render a selected dashboard.
- Render a selected plot.

The forthcoming enhancements will set the bare minimum user interactions to play with Warren.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ WARREN_API_SERVER_PORT               ?= 8100
 WARREN_FRONTEND_IMAGE_NAME           ?= warren-frontend
 WARREN_FRONTEND_IMAGE_TAG            ?= development
 WARREN_FRONTEND_IMAGE_BUILD_TARGET   ?= development
-WARREN_FRONTEND_SERVER_PORT          ?= 3000
+WARREN_FRONTEND_IMAGE_BUILD_PATH     ?= app/staticfiles/js/build/assets/index.js
 
 
 # ==============================================================================
@@ -147,7 +147,7 @@ logs: ## display frontend/api logs (follow mode)
 .PHONY: logs
 
 run: ## run the whole stack
-run: run-frontend
+run: run-app
 .PHONY: run
 
 run-app: ## run the app server (development mode)
@@ -155,6 +155,7 @@ run-app: ## run the app server (development mode)
 	@echo "Waiting for the app to be up and running..."
 	@$(COMPOSE_RUN) dockerize -wait tcp://$(DB_HOST):$(DB_PORT) -timeout 60s
 	@$(COMPOSE_RUN) dockerize -wait tcp://app:$(WARREN_APP_SERVER_PORT) -timeout 60s
+	@$(COMPOSE_RUN) dockerize -wait file:///$(WARREN_FRONTEND_IMAGE_BUILD_PATH) -timeout 60s
 .PHONY: run-app
 
 run-api: ## run the api server (development mode)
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -165,13 +166,6 @@ run-api: ## run the api server (development mode)
 	@$(COMPOSE_RUN) dockerize -wait tcp://api:$(WARREN_API_SERVER_PORT) -timeout 60s
 .PHONY: run-api
 
-run-frontend: ## run the frontend server (development mode)
-run-frontend: run-api
-	@$(COMPOSE) up -d frontend
-	@echo "Waiting for frontend to be up and running..."
-	@$(COMPOSE_RUN) dockerize -wait tcp://frontend:$(WARREN_FRONTEND_SERVER_PORT) -timeout 60s
-.PHONY: run-frontend
-
```
wilbrdt, line 174: Why not keeping the `run-frontend` rule, replacing the `dockerize -wait tcp <PORT>` with `dockerize -wait file <BUILD_FILE>`?
Now, it seems that the `app` will be build as soon as the `frontend` is up (but not necessarily ready).
Then the `app` would be ready only when the `frontend` has finished being built. Is this expected behavior?

wilbrdt (commented):
Alt text

src/frontend/apps/web/index.html
```diff
@@ -7,7 +7,8 @@
     <title>Warren</title>
   </head>
   <body>
-    <div id="root"></div>
+    <div id="warren-frontend-data" data-context="{}"></div>
```
kernicPanel, line 10: FMI, in which case is this file used ? And how data-context will be populated ?

kernicPanel (approved):
Nice work.
IMHO this projects needs tests, and maybe there are plans to adding them.
Please keep it mind that adding tests after adding code will surely leads to missing some of them.

Makefile
lebaudantoine: Actually, that would make much more sense !

src/frontend/apps/web/index.html
lebaudantoine: In terms of developer experience, I plan to add a way to mock data context while running the `vite dev` server. Hot reloading is much faster, and sometimes errors a bit more explicit. I removed the command from the package.json, to avoid any confusion. Do you want me to remove this file?

I haven't yet designed the "development" mode relying only on the frontend server, which would bypass the Django server. I'll keep you posted.

lebaudantoine:
@kernicPanel I totally agree with you. I am trying to bootstrap quickly the LTI project, and then focused mostly on connecting parts altogether. Please be sure, I'll give an extra focus in the coming weeks to make the codebase stable and reliable, by adding testing.

PR #70, opened 2023-08-17
📌(frontend) pin all dependencies being used to ensure certainty and clear visibility 

lebaudantoine:
## Proposal

To ensure absolute certainty and clear visibility of the dependencies being used, all dependencies have been pinned to their currently installed versions. This approach empowers developers with precise knowledge about the exact versions of each dependency that are in use at any given time.

For detailed insights into the advantages of employing pinned dependencies, please refer to the documentation provided by Renovate: [Dependency Pinning in Renovate](https://docs.renovatebot.com/dependency-pinning).

Special thanks to @jbpenrath for their suggestion.

## Proposition

As part of this proposal, updates were applied to three `package.json` files:

- [x] apps/web
- [x] packages/ui
- [x] at the frontend root.

### DIFF
--- a/src/frontend/apps/web/package.json
+++ b/src/frontend/apps/web/package.json
@@ -9,22 +9,22 @@
     "build-theme": "cunningham -g css -o pages"
   },
   "dependencies": {
-    "@openfun/cunningham-react": "^0.11.0",
+    "@openfun/cunningham-react": "0.11.1",
     "@vitejs/plugin-react": "4.0.3",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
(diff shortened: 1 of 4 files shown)


cyrillay (approved):

wilbrdt:
For now I guess it's fine. But what about the future, when other project will use Warren frontend as a lib? https://docs.renovatebot.com/dependency-pinning/#ranges-for-libraries

lebaudantoine:
Agree on that @wilbrdt, However I prefer to pin all dependencies for now, and make it evolve when making the UI lib public. For now, the package is explicitly private !

PR #72, opened 2023-08-17
🩹(api) fix client error handling with HttpException

lebaudantoine:
## Purpose

Use the default `HttpException` from `fastAPI` to enhance our API's error handling capabilities.

## Proposal

Integrated the default HttpException to enhance our API's error handling capabilities. The HttpException behaves like a standard Python exception with additional data relevant for APIs.

When this exception is raised, it effectively prevents the further execution of the API call. We unlock the potential for more advanced configuration options during exception handling.

Moreover, this approach eliminates the need for custom Pydantic models dedicated to managing API responses, leading to a  more maintainable code.

### DIFF
--- a/src/api/plugins/video/warren_video/api.py
+++ b/src/api/plugins/video/warren_video/api.py
@@ -1,12 +1,12 @@
 """Warren API v1 video router."""
 import logging
 
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, HTTPException
 from typing_extensions import Annotated  # python <3.9 compat
 from warren.backends import lrs_client
 from warren.fields import IRI
 from warren.filters import BaseQueryFilters, DatetimeRange
-from warren.models import DailyCounts, Error, Response, StatusEnum
+from warren.models import DailyCounts
 from warren_video.indicators import DailyCompletedViews, DailyDownloads, DailyViews
 
 router = APIRouter(
@@ -22,7 +22,7 @@ async def views(
     filters: Annotated[BaseQueryFilters, Depends()],
     complete: bool = False,
     unique: bool = False,
-) -> Response[DailyCounts]:
+) -> DailyCounts:
     """Number of views for `video_id` in the `since` -> `until` date range."""
     indicator_kwargs = {
         "client": lrs_client,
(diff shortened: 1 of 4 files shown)


lebaudantoine:
I have kept the same error "details" @cyrillay, however we may discuss it. What about passing a more predictable message instead of casting the caught exception to a string? It's quite hard for a developer to guess what would be the string passed to the client, plus searching for the string to find the place the error took place won't be possible. WDYT? 

Do you have any opinion @wilbrdt based on Ralph standards?

src/api/plugins/video/warren_video/api.py
```diff
@@ -37,24 +37,17 @@ async def views(
...
-        )
+        return await indicator.compute()
     except (KeyError, AttributeError) as exception:
-        logger.error(exception)
-        response = Response[Error](
-            status=StatusEnum.FAILED, content=Error(error_message=str(exception))
-        )
-
-    return response
+        raise HTTPException(status_code=500, detail=str(exception)) from exception
```
wilbrdt, line 42: We should still log the error for debugging purpose

src/api/plugins/video/warren_video/api.py
```diff
@@ -64,13 +57,6 @@ async def downloads(
...
```
wilbrdt, line 62: Ditto

src/api/plugins/video/warren_video/api.py
wilbrdt: Indeed @lebaudantoine we should add a specific error message, both for the logger and for the exception raised: it's easier to debug, and more explicit for the user

wilbrdt (commented):
One remark on logging the errors and good to go! :tada:

cyrillay (approved):

src/api/plugins/video/warren_video/api.py
lebaudantoine: yes good catch, my bad !

lebaudantoine: Agree on the specific error message.

src/api/plugins/video/warren_video/api.py
```diff
@@ -37,24 +37,18 @@ async def views(
...
+        return await indicator.compute()
     except (KeyError, AttributeError) as exception:
-        logger.error(exception)
-        response = Response[Error](
-            status=StatusEnum.FAILED, content=Error(error_message=str(exception))
-        )
-
-    return response
+        logger.error("An error occurred while computing the indicator: %s", exception)
+        raise HTTPException(status_code=500, detail="An error occurred while computing the indicator") from exception
```
wilbrdt, line 43:
```suggestion
        msg = "An error occurred while computing the number of views"
        logger.error("%s: %s", msg, exception)
        raise HTTPException(status_code=500, detail=msg) from exception
```
WDYT?

PR #76, opened 2023-08-23
🚧 (api) Video : Completion threshold, Name, and Length

lebaudantoine:
## Purpose

Exposes metadata of a given video resource.

## Proposal

Create a new endpoint `/info`, that returns video's metadata from `Initialized` statements.

- [X] Create a new Pydantic model to represent these data.
- [X] Create a new Indicator (name and inheritance to be discussed).
- [ ] Fetch relevant Initialized statements (still need to put a limit on the max number of statements queried).
- [ ] Filter statements to be discussed.
- [x] Add tests on the new endpoint.
- [x] Add documentation to the new indicator and Pydantic model.
- [ ] Improve exception handling on the new endpoint.

### DIFF
--- a/src/api/plugins/video/warren_video/factories.py
+++ b/src/api/plugins/video/warren_video/factories.py
@@ -124,7 +124,7 @@ class VideoDownloadedFactory(BaseXapiStatementFactory):
         },
         "timestamp": "2021-12-01T08:17:47.150905+00:00",
     }
-    model: VideoPlayed = VideoDownloaded
+    model: VideoDownloaded = VideoDownloaded
 
 
 class VideoInitializedFactory(BaseXapiStatementFactory):
--- a/src/api/plugins/video/warren_video/factories.py
+++ b/src/api/plugins/video/warren_video/factories.py
@@ -1,6 +1,10 @@
 """Factories for video xAPI events."""
 from ralph.models.xapi.concepts.constants.video import RESULT_EXTENSION_TIME
-from ralph.models.xapi.video.statements import VideoDownloaded, VideoPlayed
+from ralph.models.xapi.video.statements import (
+    VideoDownloaded,
+    VideoInitialized,
+    VideoPlayed,
+)
 from warren.factories.base import BaseXapiStatementFactory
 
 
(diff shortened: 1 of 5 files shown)


src/api/plugins/video/warren_video/models.py
```diff
@@ -1 +1,26 @@
...
+from pydantic import BaseModel, validator
+
+Seconds = int | None
+Percent = float | None
+
+
+class Info(BaseModel):
+    name: str | None
+    length: Seconds
+    completion_threshold: Percent
```
jmaupetit, line 12: Why not using simple types?

```suggestion
    length: int
    completion_threshold: float
```

Also: are we sure we want all fields to be nullable?

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -158,3 +166,72 @@ class DailyDownloads(BaseDailyEvent):
...
+
+        self.parse_raw_statements()
+
+        def get_most_common_value(column_name):
+            try:
+                values = self.statements[column_name].dropna()
+                if values.empty:
+                    return None
+                # Returning the first value is arbitrary.
+                return values.mode().iloc[0]
```
jmaupetit, line 223: makes sens to me

src/api/plugins/video/warren_video/indicators.py
```diff
@@ -158,3 +166,72 @@ class DailyDownloads(BaseDailyEvent):
...
+        self.raw_statements = [
+            value
+            async for value in self.client.read(
+                target=self.client.statements_endpoint, query=self.get_lrs_query()
+            )
+        ]
+
+    # todo - discuss whether it makes sense, as we have already loaded all statements.
+    def filter_statements(self) -> None:
+        self.raw_statements = self.raw_statements.sample(3)
```
jmaupetit, line 203: Yup. We need a way to prevent following pagination in Ralph's LRS backends.

src/api/plugins/video/warren_video/models.py
```diff
@@ -1 +1,26 @@
 """Defines the schema of what is returned inside the API responses."""
+
+from pydantic import BaseModel, validator
+
+Seconds = int | None
+Percent = float | None
+
+
+class Info(BaseModel):
+    name: str | None
```
wilbrdt, line 10:
```suggestion
    name: Union[str, None]
```
Available only from Python version 3.10+

wilbrdt (commented):
Promising!

src/api/plugins/video/warren_video/indicators.py
lebaudantoine: I will let a comment for future myself, as it's an arbitrary decision.

src/api/plugins/video/warren_video/models.py
lebaudantoine: In terms of readability, having to guess the unit of a variable seems pretty annoying for future devs. I am totally open to discuss it. Type alias also offered me to factorize the Union of the nullable and simple types, avoiding repeating it.

src/api/plugins/video/warren_video/indicators.py
lebaudantoine: It's related to this [issue](https://github.com/openfun/ralph/issues/418). Let's discuss it on Monday with @Leobouloc

src/api/plugins/video/warren_video/models.py
lebaudantoine: good catch, fixed up !

src/api/plugins/video/warren_video/factories.py
```diff
@@ -121,3 +125,63 @@ class VideoDownloadedFactory(BaseXapiStatementFactory):
         "timestamp": "2021-12-01T08:17:47.150905+00:00",
     }
     model: VideoPlayed = VideoDownloaded
```
quitterie-lcs, line 127:
```suggestion
    model: VideoDownloaded = VideoDownloaded
```

src/api/plugins/video/warren_video/indicators.py
quitterie-lcs: On Potsie, we use to return the maximum value for Completion Threshold

src/api/plugins/video/warren_video/models.py
```diff
@@ -1 +1,31 @@
...
+            raise ValueError("must be positive.")
+        return v
+
+    @validator("completion_threshold")
+    @classmethod
+    def completion_threshold_must_be_valid(cls, v: Percent) -> Percent:
+        """Validate video's completion threshold."""
+        if v is not None and (v < 0 or v > 1):
+            raise ValueError("must be between 0 and 1.")
+        return v
```
quitterie-lcs, line 31: You can use custom types to avoid define validators as it is very common validation.

Seconds could be defined as follows: https://docs.pydantic.dev/latest/api/types/#pydantic.types.PositiveInt
```python
Seconds: Union[PositiveInt, None]
```
And Percent: https://docs.pydantic.dev/latest/api/types/#pydantic.types.confloat
```python
Percent: Union[confloat(gt=0.0, lt=1.0), None]
```

src/api/plugins/video/warren_video/models.py
lebaudantoine: I am now using simple types, and added an explicit comment. WDYT?

PR #85, opened 2023-09-06
♻️(api) refactor indicators preprocessing

jmaupetit:
## Purpose

Data processing functions should not rely on a class instance attribute to apply processing on.

## Proposal

It's way more explicit to inject data to process as an argument and return processed data.


src/api/core/warren/indicators/base.py
```diff
@@ -19,23 +19,19 @@ class PreprocessMixin(object):
...
+    def add_date_column(self, statements: pd.DataFrame) -> pd.DataFrame:
         """Add a 'date' column computed from statement's timestamp."""
         # Disable chained assignment warning to make the transformation inplace
         with pd.option_context("mode.chained_assignment", None):
             # Transform 'timestamp' column into a date with the format (YYYY-MM-DD)
-            self.statements["date"] = pd.to_datetime(
-                self.statements["timestamp"]
-            ).dt.date
+            statements["date"] = pd.to_datetime(statements["timestamp"]).dt.date
+        return statements.copy()
```
lebaudantoine, line 32: Here's a quick question: should it occur before modifying the `statements`?

src/api/core/warren/indicators/base.py
```diff
@@ -44,9 +40,7 @@ def add_actor_uid_column(self) -> None:
         without worrying about the 4 IFIs.
         """
         xapi_actor_identifier_columns = (
-            settings.XAPI_ACTOR_IDENTIFIER_PATHS.intersection(
-                set(self.statements.columns)
-            )
+            settings.XAPI_ACTOR_IDENTIFIER_PATHS.intersection(set(statements.columns))
```
lebaudantoine, line 43:
```suggestion
            settings.XAPI_ACTOR_IDENTIFIER_PATHS.intersection(statements.columns)
```

src/api/core/warren/indicators/base.py
```diff
@@ -61,13 +55,16 @@ def get_uid(row):
...
+    def preprocess_statements(self, statements: List[XAPI_STATEMENT]) -> pd.DataFrame:
         """Denormalize raw statements, and add utility columns."""
-        self.parse_raw_statements()
-        self.add_actor_uid_column()
-        self.add_date_column()
+        return compose(
+            self.add_date_column,
+            self.add_actor_uid_column,
+            self.normalize_statements,
+        )(statements)
```
lebaudantoine, line 67: This lacks clarity. Without delving into the compose documentation, it appears that the steps are applied in reverse order.

src/api/core/warren/indicators/base.py
```diff
@@ -78,16 +75,23 @@ class BaseIndicator(ABC):
...
-    async def fetch_statements(self) -> None:
-        """Execute the LRS query to obtain statements required for the indicator."""
+    async def fetch_statements(
+        self, lrs_query: LRSStatementsQuery = None
+    ) -> List[XAPI_STATEMENT]:
+        """Execute the LRS query to obtain statements required for this indicator.
 
-    @abstractmethod
-    def filter_statements(self) -> None:
-        """Filter the statements required for the indicator."""
```
lebaudantoine, line 90: Nice 👍

src/api/core/warren/utils.py
```diff
@@ -0,0 +1,10 @@
+"""Warren utils."""
+from functools import reduce
+
+
+def compose(*functions):
```
lebaudantoine, line 5: Nit:  could we have type hint?

src/api/core/warren/utils.py
```diff
@@ -0,0 +1,10 @@
...
+    """Function composition.
+
+    Note that functions are applied in reverse order (last one first).
+    """
+    return reduce(lambda f, g: lambda x: f(g(x)), functions, lambda x: x)
```
lebaudantoine, line 10: Could we try calling`functions.reverse()` before 'head' to avoid the reverse order? I haven't tested it; it's up to you.

src/api/plugins/video/warren_video/api.py
```diff
@@ -28,7 +28,7 @@ async def views(
         "client": lrs_client,
         "video_id": video_id,
         "date_range": DatetimeRange.parse_obj(filters),
-        "remove_duplicate_actors": unique,
+        "unique": unique,
```
lebaudantoine, line 31: Nit: old debate, IMO unique is not explicit enough, as by definition each statement is already unique. But I agree, it feels weird to have a `unique` parameter that set a `remove_duplicate_actores` value.

lebaudantoine (commented):
🤖🤗

src/api/core/warren/utils.py
jmaupetit: The idea is to read them as you would declare them in a composition function series, _e.g._  `f(g(x))`.

src/api/core/warren/indicators/base.py
jmaupetit: Yup. That's on purpose (see my other comment). I am surprised that it's not integrated in Python stdlib. :man_shrugging:

src/api/core/warren/indicators/base.py
jmaupetit: Good point! This is silly, we should work on a copy.

src/api/core/warren/utils.py
lebaudantoine: Why did you choose this approach over a sequential order? I'm more accustomed to a sequential, step-by-step, or even a 'pipeline' approach when dealing with consecutive data modifications. I draw inspiration from this [scikit-learn class](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html).

src/api/plugins/video/warren_video/api.py
lebaudantoine: Alright your call !

src/api/core/warren/utils.py
jmaupetit: To improve code readability, I've defined a `pipe` function instead that reverse functions application order. I think it's more explicit now.

src/api/plugins/video/warren_video/api.py
```diff
@@ -25,10 +24,9 @@ async def views(
 ) -> DailyCounts:
     """Number of views for `video_id` in the `since` -> `until` date range."""
     indicator_kwargs = {
-        "client": lrs_client,
```
lebaudantoine, line 28: nice I like it

src/api/core/warren/xapi.py
```diff
@@ -0,0 +1,72 @@
...
+            ).hexdigest()
+
+        statements["actor.uid"] = statements.apply(get_uid, axis=1)
+        return statements
+
+    @staticmethod
+    def preprocess(statements: List[XAPI_STATEMENT]) -> pd.DataFrame:
+        """Denormalize raw statements, and add utility columns."""
+        return pipe(
+            StatementsTransformer.normalize_statements,
```
lebaudantoine, line 69: Nit: What do you mean by `Denormalize` ? 
Should it be:
```suggestion
        """Normalize raw statements, and add utility columns."""
        return pipe(
            StatementsTransformer.normalize_statements,
```
To be consistent with method's naming.

src/api/core/warren/tests/test_xapi.py
```diff
@@ -0,0 +1,198 @@
...
+    raw_statements = [
+        BaseXapiStatementFactory.build(
+            mutations=[{"timestamp": "2023-01-01T00:10:00.000000+00:00"}]
+        ).dict(),
+        BaseXapiStatementFactory.build(
+            mutations=[{"timestamp": "2023-01-03T00:10:00.000000+00:00"}]
+        ).dict(),
+    ]
+
+    statements = StatementsTransformer.normalize_statements(raw_statements)
```
lebaudantoine, line 36: Nit: Reading this method's call, is there any of the following options do you think is better:

- `StatementsTransformer.normalize(raw_statements)`
- `StatementsTransformer.normalize_raw(raw_statements)`

lebaudantoine (approved):
LGTM, thanks for this refactoring 👏

src/api/core/warren/xapi.py
jmaupetit: TBH I didn't noticed the docstring. :sweat_smile:

src/api/core/warren/tests/test_xapi.py
jmaupetit: Good catch! `normalize()` would be a good candidate!

src/api/core/warren/indicators/base.py
```diff
@@ -77,17 +17,32 @@ class BaseIndicator(ABC):
...
+        try:
+            return [
+                value
+                async for value in self.lrs_client.read(
+                    target=self.lrs_client.statements_endpoint,
+                    query=lrs_query or self.get_lrs_query(),
+                )
+            ]
+        except BackendException as exception:
+            raise LrsClientException("Failed to fetch statements") from exception
```
jmaupetit, line 45: We should add tests for this.

lebaudantoine (approved):

PR #87, opened 2023-09-06
🥅 (api) catch BackendException from the LRS client

lebaudantoine:
LRS client exceptions were not being handled correctly, resulting in API server failures.

To fix this issue, we created a custom `ClientException` class, and implement error handling for it in all API routes responsible for computing indicators by fetching statements.

### DIFF
--- a/src/api/plugins/video/warren_video/indicators.py
+++ b/src/api/plugins/video/warren_video/indicators.py
@@ -3,10 +3,12 @@
 
 from ralph.backends.http import BaseHTTP
 from ralph.backends.http.async_lrs import LRSQuery
+from ralph.exceptions import BackendException
 from ralph.models.xapi.concepts.constants.video import RESULT_EXTENSION_TIME
 from ralph.models.xapi.concepts.verbs.scorm_profile import CompletedVerb
 from ralph.models.xapi.concepts.verbs.tincan_vocabulary import DownloadedVerb
 from ralph.models.xapi.concepts.verbs.video import PlayedVerb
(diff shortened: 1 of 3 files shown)


src/api/plugins/video/warren_video/indicators.py
```diff
@@ -60,12 +62,15 @@ def get_lrs_query(
...
-        ]
+        try:
+            self.raw_statements = [
+                value
+                async for value in self.client.read(
+                    target=self.client.statements_endpoint, query=self.get_lrs_query()
+                )
+            ]
+        except BackendException as exception:
+            raise ClientException("Failed to fetch statements.") from exception
```
jmaupetit, line 73:
```suggestion
            raise ClientException("Failed to fetch statements") from exception
```

src/api/core/warren/backends.py
```diff
@@ -13,3 +13,7 @@
         X_EXPERIENCE_API_VERSION="1.0.3", CONTENT_TYPE="application/json"
     ),
 )
+
+
+class ClientException(Exception):
+    """Raised when the client has a failure."""
```
jmaupetit, line 19: We usually define an `exceptions.py` module (e.g. `warren.exceptions`) and define all exceptions there.

```suggestion
class LRSClientException(Exception):
    """Raised when the LRS client has a failure."""
```

wilbrdt (approved):
Looks good to me! :tada:

jmaupetit (approved):

PR #96, opened 2023-09-18
🎨(api) import FastAPI HTTP status constants to avoid hardcoding integer values

lebaudantoine:
`HTTPException` statuses were previously hard-coded with integer values. Importing Fast API constants, we not only enhance the maintainability of this logic, but also provide explicit names for each status code. Under the hood, These constants are imported from the `starlette` package. Please find all constants [here](https://github.com/encode/starlette/blob/master/starlette/status.py)

### DIFF
--- a/src/api/core/warren/filters.py
+++ b/src/api/core/warren/filters.py
@@ -1,7 +1,7 @@
 """Warren API filters."""
 
 import arrow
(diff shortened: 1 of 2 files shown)


quitterie-lcs (approved):
LGTM !

PR #99, opened 2023-09-20
⚗️(notebook) explore statements extracted from Moodle 

lebaudantoine:
## Purpose

Explore statements extracted from a Moodle instance

## Proposal

Create a first draft, of an EDA notebook. I used Plotly to make interactive plot.
Please unzip the dataset, and move the Json file to the `data`folder.

https://github.com/openfun/warren/assets/45729124/9c039cb5-ec66-4334-8b6b-ea326297c3c3

- [X] Fix Docker User while running the notebook service (thx @jmaupetit)
- [X] Pin Jupyter Docker image
- [X] Enhance performances while mounting Docker volume

## Next steps

- [ ] Select a subset of courses and students and conduct in-depth studies.
- [ ] Extract patterns from statements
- [ ] Create a multi-line plot visualizing statements for each verb within a given course

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -337,10 +337,10 @@ help:
 .PHONY: help
 
 jupytext--to-md: ## convert local ipynb files into md
-	bin/jupytext --to md work/**/*.ipynb
+	bin/jupytext --to md work/*.ipynb
 .PHONY: jupytext--to-md
 
 jupytext--to-ipynb: ## convert remote md files into ipynb
-	bin/jupytext --to ipynb work/**/*.md
+	bin/jupytext --to ipynb work/*.md
 .PHONY: jupytext--to-ipynb
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -129,4 +129,4 @@ services:
       - users
     volumes:
       - ./notebook/notebooks:/home/jovyan/work
-      - .:/home/jovyan/work/warren
+      - ./src/api:/home/jovyan/work/warren
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -130,3 +130,4 @@ services:
     volumes:
       - ./notebook/notebooks:/home/jovyan/work
       - ./src/api:/home/jovyan/work/warren
+      - ./data:/home/jovyan/work/data
--- a/notebook/Dockerfile
+++ b/notebook/Dockerfile
@@ -1,4 +1,4 @@
-FROM jupyter/base-notebook
+FROM jupyter/scipy-notebook:2023-09-25
 
 RUN pip install --upgrade pip
 
--- a/notebook/Dockerfile
+++ b/notebook/Dockerfile
@@ -3,3 +3,5 @@ FROM jupyter/scipy-notebook:2023-09-25
 RUN pip install --upgrade pip
 
 RUN pip install jupytext --upgrade
+
+RUN pip install plotly --upgrade
--- a/notebook/notebooks/2023-09-20-al-explore-moodle-data.md
+++ b/notebook/notebooks/2023-09-20-al-explore-moodle-data.md
@@ -1,0 +1,446 @@
+---
+jupyter:
+  jupytext:
+    text_representation:
+      extension: .md
+      format_name: markdown
+      format_version: '1.3'
+      jupytext_version: 1.15.2
+  kernelspec:
+    display_name: Python 3 (ipykernel)
+    language: python
+    name: python3
+---
+
+```python
+!pip install itables > /dev/null
+```
+
+<!-- #region jupyter={"outputs_hidden": false} -->
+_______________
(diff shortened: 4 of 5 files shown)


notebook/notebooks/2023-09-20-al-explore-moodle-data.md
```diff
@@ -0,0 +1,451 @@
...
+      format_version: '1.3'
+      jupytext_version: 1.15.1
+  kernelspec:
+    display_name: Python 3 (ipykernel)
+    language: python
+    name: python3
+---
+
+```python
+!pip install pandas numpy matplotlib itables >> null
```
jmaupetit, line 16: You have the `/dev/null` device for that. :wink: 

```suggestion
!pip install pandas numpy matplotlib itables > /dev/null
```

notebook/Dockerfile
```diff
@@ -1,5 +1,8 @@
-FROM jupyter/base-notebook
+FROM jupyter/base-notebook:2023-09-11
 
 RUN pip install --upgrade pip
 
 RUN pip install jupytext --upgrade
+
+RUN pip install ipywidgets>=7.6 --upgrade
+RUN pip install plotly --upgrade
```
jmaupetit, line 8: I think we rather should install them in the notebook that requires them. WDYT?

jmaupetit (approved):
I only have minor comments. That's a rather complete study :1st_place_medal:

docker-compose.yml
```diff
@@ -117,7 +117,7 @@ services:
   notebook:
     build:
       context: ./notebook
-    user: ${DOCKER_USER:-1000}
+    user: "root:root"
```
quitterie-lcs, line 120: I would recommend not to change it

notebook/Dockerfile
quitterie-lcs: Maybe we could switch to `datascience-notebook` or an other more specific image than the `base` one

notebook/notebooks/2023-09-20-al-explore-moodle-data.md
lebaudantoine: indeed, that was I had in mind

notebook/Dockerfile
lebaudantoine: Actually, I tried, however, plotly needs to be installed before starting the Jupyter lab server.

docker-compose.yml
lebaudantoine: @jmaupetit suggested me this change, let's discuss it during a stand-up.

notebook/notebooks/2023-09-20-al-explore-moodle-data.md
```diff
@@ -0,0 +1,451 @@
...
+filepath = Path('./data') / filename
+```
+
+```python jupyter={"outputs_hidden": false} pycharm={"name": "#%%\n"}
+%%time
+# Open file's content
+lines = []
+with open(filepath) as file:
+    for line in file.readlines():
+        lines.append(line.rstrip())
```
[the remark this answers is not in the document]
lebaudantoine: Hum, it would load the file's content in a single string. How would you convert it to JSON? The JSON file is not a valid JSON array.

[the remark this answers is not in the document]
jmaupetit: Sorry I forgot the `split()` to do the exact same thing as the original with/open/for loop (no stripping):

```suggestion
lines = filepath.read_text().split()
```

notebook/Dockerfile
jmaupetit: That's too bad.

lebaudantoine: I have switched to `datascience-scipy` to have a more 'scientific' image.

Makefile
```diff
@@ -337,10 +337,10 @@ help:
 .PHONY: help
 
 jupytext--to-md: ## convert local ipynb files into md
-	bin/jupytext --to md work/**/*.ipynb
+	bin/jupytext --to md work/*.ipynb
 .PHONY: jupytext--to-md
 
 jupytext--to-ipynb: ## convert remote md files into ipynb
-	bin/jupytext --to ipynb work/**/*.md
+	bin/jupytext --to ipynb work/*.md
```
lebaudantoine, line 344: I've updated the regex to avoid matching .md files like LICENSE or README in the 'warren' shared volume. If you know any clever tricks to exclude a subdirectory, 'work' in this case, with Jupytext CLI, please share. My attempts were not successful.

Currently, we have just a few notebooks and no complex folder structure to worry about.

notebook/Dockerfile
jmaupetit: I do agree! Let's keep it this way.

PR #100, opened 2023-09-22
👷(circle) move to a single docker image name

jmaupetit:
## Purpose

All docker images required to run Warren will be called `warren` and we distinguish application services by their image tag.

## Proposal

- `warren:app-latest` is the latest warren app service
- `warren:api-core-latest` is the latest warren api service (core version)
- `warren:api-full-latest` is the latest warren api service (full version, i.e. with all official plugins installed)

`latest` points to the latest release, but each version tag will also be published (e.g. `warren:app-0.1.0`, `warren:api-core-0.1.0`, `warren:api-full-0.1.0`).

Docker images repository: https://hub.docker.com/repository/docker/fundocker/warren/general

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,14 +119,20 @@ jobs:
       # Each image is tagged with the current git commit sha1 to avoid
       # collisions in parallel builds.
       - run:
-          name: Build production image
+          name: Build core production image
           command: |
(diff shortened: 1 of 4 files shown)


wilbrdt (approved):
Alt text
LGTM!

PR #119, opened 2023-10-20
✨(frontend) upgrade Header with loading status

lebaudantoine:
## Purpose

Reduce redundancy of loading spinners across page components.

## Proposal

Enhance the header component with:

- A logo
- A breadcrumb displaying Course's info from LTI context
- A loading bar to reflect data fetching status

https://github.com/openfun/warren/assets/45729124/4e76f696-281a-456d-8b83-696074034411

To do:
- [x] Update `LTIRequestView` tests

Next Steps:
- [ ] Update the current Breadcrumb component with Cunningham's one
- [ ] Add functionalities in the Header menu.


src/frontend/packages/ui/types/index.ts
```diff
@@ -1,7 +1,15 @@
+export interface CourseData {
+  school_name: string;
+  course_name: string;
+  course_run?: string;
```
jmaupetit, line 4: This is LMS-specific (edx mostly). Maybe we should be more generic using:

```suggestion
  platform: string;
  organization?: string;
  course_name: string;
  course_run?: string;
```
WDYT?

jmaupetit (commented):
Looks amazing! :muscle:

src/frontend/packages/ui/types/index.ts
lebaudantoine: Agree, It would ask a bit of data manipulation, but make the code more generic

lebaudantoine: I have renamed it in the back-end, WDYT?

jmaupetit: Looks great! We need to keep in mind multi-tenancy (_i.e._ multiple platforms support).

src/frontend/packages/ui/utils.scss
```diff
@@ -0,0 +1,18 @@
+@use "sass:math";
```
jbpenrath, line 1: Do not hesitate to open an issue on https://github.com/openfun/cunningham to discuss about that.

src/frontend/packages/ui/components/LoadingBarWrapper/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  useEffect(() => {
+    if (initialFetchCount === 0 && currentFetchCount) {
+      setInitialFetchCount(currentFetchCount);
+      setLoadingProgress(INITIAL_PROGRESS);
+    } else if (initialFetchCount !== 0) {
+      setLoadingProgress(
+        (100 / initialFetchCount) * (initialFetchCount - currentFetchCount),
+      );
+    }
+  }, [currentFetchCount]);
```
jbpenrath, line 36: I wonder if you could not save a render loop by setting directly currentFetchCount as the default value of the state ?

```tsx
  const currentFetchCount = useIsFetching();
  const [initialFetchCount, setInitialFetchCount] = useState(currentFetchCount);
```

src/frontend/packages/ui/components/LoadingBarWrapper/_index.scss
```diff
@@ -0,0 +1,12 @@
+
+.c__loading-bar {
+  // FIXME - Centralize and reorganize variables for better maintainability.
+  $body-padding: 8px;
+  &__container {
```
jbpenrath, line 5: No way to use an internal class of loading bar to increase selector specifity instead of using `!important` ?

```scss
&__container.internal-loading-bar-class {
```

src/frontend/packages/ui/components/Header/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const Header: React.FC = () => {
+  const { appData } = useLTIContext();
+  return (
+    <>
+      <header className="c__header">
+        <div className="c__header__logo-title">
+          <Logo />
+          <h1>Warren</h1>
+        </div>
+        <div className="c__header__separator" />
```
jbpenrath, line 26: I wonder if this separator could be done only with CSS through pseudo element. I do not the full context so surely this is more relavant to do that through a DOM element.

src/frontend/packages/ui/utils.scss
lebaudantoine: I didn't, as it has already been fixed by https://github.com/openfun/cunningham/pull/195

src/frontend/packages/ui/components/LoadingBarWrapper/_index.scss
lebaudantoine: I tried, but they didn't use any CSS class internally. They directly styled element...

src/frontend/packages/ui/components/Header/index.tsx
lebaudantoine: good suggestion 🤗 I'll fix it

PR #133, opened 2023-11-13
🐛(api) fix views computation

quitterie-lcs:
## Purpose 

Views is computed by counting `played` events generated before a threshold
on the video timeline.

## Proposal

Currently, indicators based on views computation was counting events generated
after this threshold. It has been reversed and tests adapted.


lebaudantoine (approved):
LGTM

jmaupetit (approved):

PR #134, opened 2023-11-13
🔊 (api) enhance logging strategy

lebaudantoine:
## Purpose

Optimize logging levels for improved clarity and efficiency.

## Proposal

- [x] Add colors on Uvicorn logs.
- [x] Update Uvicorn logs to include core and video APIs with DEBUG level.
- [x] Add some INFO logs in video's endpoints.
- [x] Make debug logs more explicit and readable.
- [x] Use `logging.exception` instead of `logging.error`.

### DIFF
--- a/src/api/core/logging-config.yaml
+++ b/src/api/core/logging-config.yaml
@@ -32,7 +32,12 @@ loggers:
       - access
     level: INFO
     propagate: false
-  warren.api:
+  core.warren.api.v1:
+    handlers:
+      - default
+    level: INFO
+    propagate: false
+  warren_video.api:
     handlers:
       - default
     level: INFO
--- a/src/api/core/logging-config.yaml
+++ b/src/api/core/logging-config.yaml
@@ -5,10 +5,11 @@ formatters:
   default:
     "()": uvicorn.logging.DefaultFormatter
     fmt: "%(levelprefix)s %(message)s"
-    use_colors:
+    use_colors: true
   access:
     "()": uvicorn.logging.AccessFormatter
     fmt: '%(levelprefix)s %(client_addr)s - "%(request_line)s" %(status_code)s'
+    use_colors: true
 handlers:
   default:
     formatter: default
--- a/src/api/core/warren/indicators/models.py
+++ b/src/api/core/warren/indicators/models.py
@@ -11,7 +11,7 @@ from sqlmodel import Field, SQLModel
 
 
 class CacheEntryCreate(SQLModel):
-    """Indicator generic persistency."""
+    """Indicator generic persistence."""
 
     id: Optional[UUID] = Field(default_factory=lambda: uuid4().hex, primary_key=True)
     key: str = Field(max_length=100, index=True)
@@ -25,4 +25,4 @@ class CacheEntryCreate(SQLModel):
 
 
 class CacheEntry(CacheEntryCreate, table=True):
-    """Indicator generic persistency (table version)."""
+    """Indicator generic persistence (table version)."""
--- a/src/api/plugins/video/warren_video/api.py
+++ b/src/api/plugins/video/warren_video/api.py
@@ -34,6 +34,7 @@ async def views(
 ) -> DailyCounts:
     """Number of views for `video_id` in the `since` -> `until` date range."""
     # Switch/case pattern matching with the (complete, unique) boolean tuple
+    logger.debug("Start computing 'views' indicator")
     klass_mapping = {
         (True, True): DailyUniqueCompletedViews,
         (True, False): DailyCompletedViews,
@@ -43,20 +44,24 @@ async def views(
     indicator = klass_mapping.get((complete, unique), DailyViews)(
         video_id=video_id, span_range=DatetimeRange.parse_obj(filters)
     )
-    logger.debug("Will compute indicator %s", indicator)
+    logger.debug("Will compute indicator %s", type(indicator).__name__)
+    logger.debug(
+        "From %s to %s", indicator.span_range.since, indicator.span_range.until
+    )
 
     try:
         results = await indicator.get_or_compute()
     except (KeyError, AttributeError, LrsClientException) as exception:
         message = "An error occurred while computing the number of views"
-        logger.error("%s: %s", message, exception)
+        logger.exception("%s. Exception:", message)
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=message
         ) from exception
 
     if isinstance(results, DailyUniqueCounts):
         results = results.to_daily_counts()
-    logger.debug("results = %s", results)
+    logger.debug("Results = %s", results)
+    logger.debug("Finish computing 'views' indicator")
     return results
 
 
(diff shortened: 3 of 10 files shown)


src/api/core/warren/indicators/models.py
```diff
@@ -11,7 +11,7 @@
 
 
 class CacheEntryCreate(SQLModel):
-    """Indicator generic persistency."""
```
lebaudantoine, line 14: Note: "persistency" is acceptable, but not in my IDE dictionary 🥶

src/api/core/logging-config.yaml
```diff
@@ -32,8 +32,13 @@ loggers:
...
+  core.warren.api.v1:
     handlers:
       - default
-    level: INFO
+    level: DEBUG
+    propagate: false
+  warren_video.api:
+    handlers:
+      - default
+    level: DEBUG
```
lebaudantoine, line 43: These levels can be discussed !

src/api/plugins/video/warren_video/api.py
```diff
@@ -34,6 +34,7 @@ async def views(
 ) -> DailyCounts:
     """Number of views for `video_id` in the `since` -> `until` date range."""
     # Switch/case pattern matching with the (complete, unique) boolean tuple
+    logger.info("Start computing 'views' indicator")
```
lebaudantoine, line 37: It would add more complexity than necessary, but I could have to reuse the function's name instead of duplicating it. IMO the duplication is fine.

src/api/plugins/video/warren_video/api.py
```diff
@@ -43,20 +44,22 @@ async def views(
...
-        logger.error("%s: %s", message, exception)
+        logger.exception(f"{message}. Exception:")
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=message
         ) from exception
 
     if isinstance(results, DailyUniqueCounts):
         results = results.to_daily_counts()
-    logger.debug("results = %s", results)
+    logger.debug(f"Results = {results.json()}")
```
lebaudantoine, line 61: Would you prefer a `json.dumps` to pretty log these results?

src/api/plugins/video/warren_video/api.py
```diff
@@ -68,22 +71,25 @@ async def downloads(
...
```
lebaudantoine, line 93: ditto.

src/api/core/logging-config.yaml
```diff
@@ -31,8 +32,13 @@ loggers:
...
     propagate: false
```
lebaudantoine, line 44: Might not be the right place to declare this part, as it's a plugin.

src/api/core/warren/indicators/models.py
jmaupetit: Yup, I've noticed that we also can say "Persistence" in English. Feel free to change this!

src/api/core/logging-config.yaml
jmaupetit: Those are fine for development, but not suitable for  production. I would recommend to:

1. Distribute a default, production-ready basic configuration (similar to the initial configuration),
2. Make logging configurable (using a `ConfigMap` in the Helm Chart),
3. Override logging configuration in development using a docker-compose volume.

src/api/core/logging-config.yaml
jmaupetit: It's fine if you consider that this plugin is bundled with the core application in a Docker image. But it should be overridable.

src/api/plugins/video/warren_video/api.py
jmaupetit: This is not something we want to log at this level. The access logs are already here for this.

```suggestion
    logger.debug("Start computing 'views' indicator")
```

src/api/plugins/video/warren_video/api.py
```diff
@@ -43,20 +44,22 @@ async def views(
     indicator = klass_mapping.get((complete, unique), DailyViews)(
         video_id=video_id, span_range=DatetimeRange.parse_obj(filters)
     )
-    logger.debug("Will compute indicator %s", indicator)
+    logger.debug(f"Will compute indicator {type(indicator).__name__}")
+    logger.debug(f"From {indicator.span_range.since} to {indicator.span_range.until}")
```
jmaupetit, line 48: Pylint would have told you that we are not supposed to use f-strings in logs, but plain old-style logs formatting:

```suggestion
    logger.debug("Will compute indicator %s", type(indicator).__name__)
    logger.debug("From %s to %s", indicator.span_range.since, indicator.span_range.until)
```

References:

1. https://pylint.pycqa.org/en/stable/user_guide/messages/warning/logging-fstring-interpolation.html
2. https://www.reddit.com/r/learnpython/comments/xx14jw/why_not_use_fstrings_in_logging/

src/api/plugins/video/warren_video/api.py
jmaupetit: We usually log the string representation of an object, which was implicit with previous implementation.

jmaupetit (commented):
I love colored logs. :heart_eyes:

src/api/plugins/video/warren_video/api.py
lebaudantoine: Alright, gotcha. In a production setup, would the Uvicorn access logs and the logger's ones be on the same stream?

src/api/plugins/video/warren_video/api.py
lebaudantoine: The current linter triggers no warning. Alright, thanks for the references, kinda suck that we still need to use this old syntax.

lebaudantoine: fixed !

src/api/plugins/video/warren_video/api.py
lebaudantoine: DEBUG:    Start computing 'views' indicator
DEBUG:    Will compute indicator DailyUniqueViews
DEBUG:    From 2023-11-07 00:00:00+01:00 to 2023-11-14 23:59:59+01:00
DEBUG:    Results = {"total": 26, "counts": [{"date": "2023-11-07", "count": 1}, {"date": "2023-11-08", "count": 2}, {"date": "2023-11-09", "count": 4}, {"date": "2023-11-10", "count": 4}, {"date": "2023-11-11", "count": 7}, {"date": "2023-11-12", "count": 2}, {"date": "2023-11-13", "count": 6}, {"date": "2023-11-14", "count": 0}]}
DEBUG:    Finish computing 'views' indicator

lebaudantoine: The string representation of these results was quite a pain to read. We had several levels of nesting to get the info. I preferred more concise and explicit logs as follow:

lebaudantoine: Fixed !

src/api/plugins/video/warren_video/api.py
jmaupetit: In a kubernetes context, your application should log everything in standard unix streams (stdin, stdout), or dedicated logging services/tools. Apart from that, IMO this message is not relevant for a production environment: what would you expect from it?

src/api/plugins/video/warren_video/api.py
jmaupetit: IMO the second is more informative as types are explicit, which could be useful in a development environment.

src/api/core/logging-config.yaml
```diff
@@ -5,10 +5,11 @@ formatters:
   default:
     "()": uvicorn.logging.DefaultFormatter
     fmt: "%(levelprefix)s %(message)s"
-    use_colors:
+    use_colors: True
   access:
     "()": uvicorn.logging.AccessFormatter
     fmt: '%(levelprefix)s %(client_addr)s - "%(request_line)s" %(status_code)s'
+    use_colors: True
```
jmaupetit, line 12:
```suggestion
    use_colors: true
  access:
    "()": uvicorn.logging.AccessFormatter
    fmt: '%(levelprefix)s %(client_addr)s - "%(request_line)s" %(status_code)s'
    use_colors: true
```

src/api/plugins/video/warren_video/api.py
lebaudantoine: Afterthoughts, I do agree with you.

src/api/core/logging-config.yaml
lebaudantoine: done, I would need to rewrite/update some commit's. I chose to pass the log level as an option of the `uvicorn` command in the docker-compose

jmaupetit (approved):

PR #137, opened 2023-11-14
🐛(api) fix Cache mixin database session management

lebaudantoine:
## Purpose

Have a proper management of database sessions while using cache.

## Proposal

Resolved issues within Cache mixin classes where database sessions were not being properly closed, resulting in exceptions from SQLAlchemy.

FastAPI is now responsible for acquiring a single database session per request and subsequently closing it after the response is returned. This update ensures proper management of database sessions, preventing to exceed the connections pool size limitations.

In this initial approach, I passed the session as a parameter to the `get_or_compute` method. This simplification removed the Cache mixin's internal state regarding the database session. Wdyt @jmaupetit?

### DIFF
--- a/src/api/core/warren/indicators/mixins.py
+++ b/src/api/core/warren/indicators/mixins.py
@@ -10,7 +10,6 @@ import arrow
 from pydantic.main import BaseModel
 from sqlmodel import Session, select
 
-from warren.db import get_session as get_db_session
 from warren.filters import DatetimeRange
 
 from .models import CacheEntry, CacheEntryCreate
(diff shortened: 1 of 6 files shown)


src/api/core/warren/indicators/mixins.py
```diff
@@ -62,20 +56,21 @@ def cache_key(self) -> str:
         ).hexdigest()
         return f"{self.__class__.__name__.lower()}-{lrs_query_hash}"
 
-    async def get_cache(self) -> CacheEntry:
+    async def get_cache(self, session: Session) -> CacheEntry:
```
jmaupetit, line 59: I think `session` is not explicit in the indicator context. WDYT?

```suggestion
    async def get_cache(self, db_session: Session) -> CacheEntry:
```

src/api/core/warren/indicators/mixins.py
```diff
@@ -43,10 +42,7 @@
 class CacheMixin:
     """A cache mixin that handles indicator persistence."""
 
-    @cached_property
-    def db_session(self) -> Session:
-        """Get the database session singleton."""
-        return next(get_db_session())
+    db_session: Session = None
```
lebaudantoine, line 45: Error handling is missing when the `db_session` is None.

PR #141, opened 2023-11-16
🚑️(frontend) configure API and APP root urls in production

lebaudantoine:
Previously, the APIs root URLs were statically resolved during the frontend build using Vite. This meant that changing these URLs required a rebuild of the frontend image for a new user.

The frontend environment variables are designed to be overridden by values passed through the Django template. This design choice enables the specification of APIs root URLs in production without necessitating a rebuild of the frontend image.


.env.dist
```diff
@@ -30,6 +30,10 @@ WARREN_APP_REFRESH_TOKEN_LIFETIME=86400
 WARREN_APP_LTI_ACCESS_TOKEN_LIFETIME=86400
 WARREN_APP_ALLOWED_ORIGINS=http://localhost:5173
 
+# Warren Frontend
+WARREN_FRONTEND_API_ROOT_URL:=http://localhost:8100
+WARREN_FRONTEND_APP_ROOT_URL:=http://localhost:8090
```
jmaupetit, line 35:
```suggestion
WARREN_API_ROOT_URL:=http://localhost:8100
WARREN_APP_ROOT_URL:=http://localhost:8090
```

src/app/apps/lti/templates/base.html
```diff
@@ -2,6 +2,11 @@
 
 <!DOCTYPE html>
 <html>
+  <!-- Pass App and API base URLs to the frontend as global variable if specified in the .env file -->
+  <script>
+    const WARREN_FRONTEND_API_ROOT_URL = "{{ WARREN_FRONTEND_API_ROOT_URL }}";
+    const WARREN_FRONTEND_APP_ROOT_URL = "{{ WARREN_FRONTEND_APP_ROOT_URL }}";
```
jmaupetit, line 8:
```suggestion
    const WARREN_API_ROOT_URL = "{{ WARREN_API_ROOT_URL }}";
    const WARREN_APP_ROOT_URL = "{{ WARREN_APP_ROOT_URL }}";
```

src/frontend/packages/ui/libs/axios.ts
```diff
@@ -1,6 +1,8 @@
 import Axios from "axios";
 
+// Retrieve Warren API root URL from Django Template global variable or Vite environment.
 const API_BASE_URL = `${
+  WARREN_FRONTEND_API_ROOT_URL ||
```
jmaupetit, line 5:
```suggestion
  WARREN_API_ROOT_URL ||
```

src/frontend/packages/ui/libs/axios.ts
```diff
@@ -10,7 +12,9 @@ export const apiAxios = Axios.create({
   withCredentials: true,
 });
 
+// Retrieve Warren App's API root URL from Django Template global variable or Vite environment.
 const APP_BASE_URL = `${
+  WARREN_FRONTEND_APP_ROOT_URL ||
```
jmaupetit, line 17:
```suggestion
  WARREN_APP_ROOT_URL ||
```

src/app/apps/lti/views.py
```diff
@@ -89,6 +90,19 @@ def render_to_response(self, context=None, **response_kwargs):
...
+        # If not specified in the .env, frontend should declare their values.
+        context.update(
+            {
+                key: os.getenv(key, "")
+                for key in [
+                    "WARREN_FRONTEND_API_ROOT_URL",
+                    "WARREN_FRONTEND_APP_ROOT_URL",
+                ]
+            }
+        )
```
jmaupetit, line 104: A better way to handle this is to add those two variables to the settings (that can be overridden by the environment) and use them here as follow:

```suggestion
        # To put on top (if not already loaded)
        from django.conf import settings
        
        context.update(
            {
                "WARREN_API_ROOT_URL": settings.API_ROOT_URL,
                "WARREN_APP_ROOT_URL": settings.APP_ROOT_URL,
            }
        )
```

jmaupetit (approved):

PR #142, opened 2023-11-16
🐛(indicators) fix method inheritance order for views indicators

quitterie-lcs:
## Purpose 

Views indicators tests pass for all requested `played` statements.
It has been designed that normally such indicators should only compute on
`played` statements that were generated in the 30 first seconds of a video. 

## Proposal

The mixin used for DailyViews was placed after the DailyEvent class in
inheritance, causing method resolution order issues. By reordering the
inheritance sequence, indicators now correctly inherit methods from the mixin.

### DIFF
--- a/src/api/plugins/video/tests/test_api.py
+++ b/src/api/plugins/video/tests/test_api.py
@@ -236,9 +236,9 @@ async def test_unique_views_backend_query(
                 ).json(),
             )
             for view_data in [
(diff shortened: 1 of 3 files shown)


lebaudantoine (approved):
LGTM, I got a small question. Should we add more tests on this filtering feature? I will be more than happy to help you on that.

jmaupetit (approved):

quitterie-lcs:
@lebaudantoine Yes, I'll add tests in my next PR

PR #143, opened 2023-11-17
Add mypy (static type checking)

lebaudantoine:
## Purpose

Improve the code quality.

## Proposal

Mypy is a static type checker for Python that aims to combine the benefits of dynamic (or "duck") typing and static typing.
It closes #31 .

Todo
- [ ] Fix mixins related errors.
- [ ] Fix plugin architecture related errors.
- [ ] Discuss introduce change on Cache and Indicator classes.
- [ ] Update tests


src/api/core/warren/indicators/mixins.py
```diff
@@ -143,10 +145,11 @@ class IncrementalCacheMixin(CacheMixin, ABC):
 
     @staticmethod
     @abstractmethod
-    def merge(a: dict, b: dict):
+    def merge(a: Any, b: Any) -> Any:
```
quitterie-lcs, line 148: Why do you need to expand the type to `Any` here ?

src/api/core/warren/xapi.py
```diff
@@ -69,7 +69,7 @@ def preprocess(
         """Normalize raw statements, and add utility columns."""
         if statements is None or not len(statements):
             logger.info("There are no statements to process")
-            return
+            return None
```
quitterie-lcs, line 72: :medal_sports:

src/api/core/warren/indicators/mixins.py
lebaudantoine: I think it's because, this method is overridden by the children classes, which specify these types. However, a better approach might be to use a Generic type alias, and find a way to pass it, when declaring the inheritance of this mixin.

jmaupetit (commented):
Looks good to me! Next step is the app?

src/api/core/warren/db.py
```diff
@@ -9,7 +11,7 @@
 engine = create_engine(settings.DATABASE_URL, echo=settings.DEBUG)
 
 
-def get_session() -> Session:
+def get_session() -> Generator[Session, None, None]:
```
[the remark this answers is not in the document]
lebaudantoine: Please have a look [here](https://docs.python.org/3/library/typing.html?highlight=generator#typing.Generator) `Generator[yield_type, send_type, return_type]`. I could alternatively try `Iterable[Session]`, would you prefer it?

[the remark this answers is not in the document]
quitterie-lcs: Thank you, keep the Generator typing, that's fine for me

quitterie-lcs (approved):

PR #144, opened 2023-11-17
✨(api) add health check endpoints

jmaupetit:
## Purpose

When deploying an application in production (e.g. in a Kubernetes environment), we need a way to check its status so that we can restart it if required.

## Proposal

We've implemented [Mozilla's Dockerflow](https://github.com/mozilla-services/Dockerflow) endpoints.

- [x]  add `__hearbeat__` and `__lbheartbeat__` API endpoints
- [x] add tests for database status check

### DIFF
--- a/src/api/core/warren/api/health.py
+++ b/src/api/core/warren/api/health.py
@@ -1,0 +1,62 @@
+"""API routes related to application health checking."""
+
+import logging
+from enum import Enum, unique
+
+from fastapi import APIRouter, Response, status
+from pydantic import BaseModel
+from ralph.backends.http.base import HTTPBackendStatus
+
+from warren.backends import lrs_client
+from warren.db import is_alive as is_db_alive
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter()
+
+
+@unique
+class BackendStatus(str, Enum):
+    """Generic backend statuses inspired from Ralph HTTP backend."""
+
+    OK = "ok"
+    AWAY = "away"
+    ERROR = "error"
+
+
+class Heartbeat(BaseModel):
+    """Warren backends status."""
+
+    database: BackendStatus
+    lrs: HTTPBackendStatus
+
+    @property
+    def is_alive(self):
+        """A helper that checks the overall status."""
+        if self.database == BackendStatus.OK and self.lrs == HTTPBackendStatus.OK:
+            return True
+        return False
+
+
+@router.get("/__lbheartbeat__")
+async def lbheartbeat() -> None:
+    """Load balancer heartbeat.
+
+    Return a 200 when the server is running.
+    """
+    return
+
+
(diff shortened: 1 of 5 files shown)


src/api/core/warren/api/health.py
```diff
@@ -0,0 +1,40 @@
...
+    content = {
+        "database": "ok" if is_db_alive() else "error",
+        "lrs": (await lrs_client.status()).value,
+    }
+    status_code = (
+        status.HTTP_200_OK
+        if all(v == "ok" for v in content.values())
+        else status.HTTP_500_INTERNAL_SERVER_ERROR
+    )
+    return JSONResponse(status_code=status_code, content=content)
```
lebaudantoine, line 40: Wdyt of having an Enum to avoid duplicated hard-coded literals? or reuse the `HTTPBackendStatus`?

lebaudantoine (approved):
LGTM

src/api/core/warren/api/health.py
lebaudantoine: My comment might be nit-picking, but I feel a bit uncomfortable having to deal with hard-coded literals, especially when LRS status derived from a well encapsulated Enum.
Adding a new enum doesn't solve all the issue, as the enum could unsynchronized with `HTTPBackendStatus`.

lebaudantoine:
```suggestion
async def heartbeat() -> JSONResponse:
    """Application heartbeat.

    Return a 200 if all checks are successful.
    """
    content = {
        "database": HTTPBackendStatus.OK if is_db_alive() else HTTPBackendStatus.ERROR,
        "lrs": (await lrs_client.status()).value,
    }
    status_code = (
        status.HTTP_200_OK
        if all(v == HTTPBackendStatus.OK for v in content.values())
        else status.HTTP_500_INTERNAL_SERVER_ERROR
    )
    return JSONResponse(status_code=status_code, content=content)
```

jmaupetit: I get it. But we need a more generic internal Enum as `HTTPBackendStatus` doesn't seems relevant for a database.

lebaudantoine: yes totally agree !

jmaupetit: I've added a fixup commit that implements this. WDYT?

src/api/core/warren/api/health.py
```diff
@@ -13,6 +15,29 @@
 router = APIRouter()
 
 
+@unique
+class BackendStatus(Enum):
```
lebaudantoine, line 19: nit:
```suggestion
class BackendStatus(str, Enum):
```

lebaudantoine (approved):
Thanks, this is super clear. This refactoring was worth it.

PR #146, opened 2023-11-20
✅(api) add load tests for a video route

jmaupetit:
## Purpose

## Proposal

- [ ] add a test to reproduce the issue
- [ ] fix the issue

### DIFF
--- a/src/api/core/logging-config.dev.yaml
+++ b/src/api/core/logging-config.dev.yaml
@@ -37,6 +37,11 @@ loggers:
       - default
     level: DEBUG
     propagate: false
+  warren:
+    handlers:
+      - default
+    level: DEBUG
+    propagate: false
   warren_video.api:
     handlers:
       - default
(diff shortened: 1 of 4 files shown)


jmaupetit:
This work is related to #137

src/api/plugins/video/tests/test_api.py
```diff
@@ -536,3 +537,58 @@ def lrs_response(request: httpx.Request):
...
+    ]
+    httpx_mock.add_response(
+        url=re.compile(r"^http://fake-lrs\.com/xAPI/statements\?.*$"),
+        method="GET",
+        json={"statements": statements},
+        status_code=200,
+    )
+
+    # Perform the call to warren backend. When fetching the LRS statements, it will
+    # get the above mocked statements
```
jmaupetit, line 574:
```suggestion
```

jmaupetit:
### DO NOT MERGE ###

This PR is still a test and increase too much tests execution time to be worth it.

PR #149, opened 2023-11-23
🔧(app) improve settings management

jmaupetit:
## Purpose

Django settings should be easily configurable in various environments with an homogeneous naming pattern.

## Proposal

- [x] allow to configure logging
- [x] make naming pattern predictable
- [x] remove duplicates


src/app/warren/settings.py
```diff
@@ -119,20 +160,12 @@ class Base(Configuration):
 
     # Static files (CSS, JavaScript, Images)
     # https://docs.djangoproject.com/en/4.1/howto/static-files/
-    STATIC_ROOT = values.Value(
-        BASE_DIR / Path("static"),
-        environ_name="WARREN_APP_STATIC_ROOT",
-        environ_prefix=None,
-    )
+    STATIC_ROOT = Value(BASE_DIR / Path("static"))
```
lebaudantoine, line 163: How is pass the environ_name with this refacto? is this some Django magic (matching automatically the settings constant's name with an environment variable)?

src/app/warren/settings.py
```diff
@@ -262,50 +277,40 @@ class Base(Configuration):
     }
 
     LTI_ACCESS_TOKEN_LIFETIME = timedelta(
-        seconds=values.FloatValue(
-            default=86400,  # 24 hours
-            environ_name="WARREN_APP_LTI_ACCESS_TOKEN_LIFETIME",
-            environ_prefix=None,
-        )
+        seconds=FloatValue(default=86400, environ_name="LTI_ACCESS_TOKEN_LIFETIME")
```
lebaudantoine, line 280: why do you need to pass the `environ_name` here?

src/app/warren/settings.py
```diff
@@ -262,50 +277,40 @@ class Base(Configuration):
...
     # Frontend
     ROOT_URLS = {
-        "API_ROOT_URL": values.Value(
+        "API_ROOT_URL": Value(
             "", environ_name="WARREN_API_ROOT_URL", environ_prefix=None
         ),
-        "APP_ROOT_URL": values.Value(
-            "", environ_name="WARREN_APP_ROOT_URL", environ_prefix=None
-        ),
+        "APP_ROOT_URL": Value("ROOT_URL", environ_name="ROOT_URL"),
```
lebaudantoine, line 313: Some changed have been introduced ? The default value for `APP_ROOT_URL` is now `'ROOT_URL'` and for `API_ROOT_URL` still empty. Is it desired?

src/app/warren/settings.py
```diff
@@ -383,20 +384,19 @@ class ContinuousIntegration(Test):
...
 
-    You must define the DJANGO_ALLOWED_HOSTS environment variable in Production
+    You must define the WARREN_APP_ALLOWED_HOSTS environment variable in Production
     configuration (and derived configurations):
-    DJANGO_ALLOWED_HOSTS="foo.com,foo.fr".
+    WARREN_APP_ALLOWED_HOSTS="foo.com,foo.fr".
     """
 
     # Security
-    ALLOWED_HOSTS = values.ListValue(None)
```
lebaudantoine, line 392: Do we accept by default all hosts? Was it redundant with `CORS_ALLOWED_ORIGINS`?

lebaudantoine (commented):
few questions, it's amazingly clear, thx again, I love the factorization 🤯

src/app/warren/settings.py
jmaupetit: Yes, Django configurations automagically assign the corresponding environment variable to this setting given the setting name.

src/app/warren/settings.py
jmaupetit: Because it cannot be auto-magically discovered as it's not assigned to a setting variable.

src/app/warren/settings.py
jmaupetit: Well spotted!

src/app/warren/settings.py
jmaupetit: > Do we accept by default all hosts?

Only in `Development`.

> Was it redundant with CORS_ALLOWED_ORIGINS?

No. `ALLOWED_HOSTS` refers to the website domain name serving the application. `CORS_ALLOWED_ORIGINS` may be larger than that.

lebaudantoine (approved):
LGTM

PR #153, opened 2023-11-28
Fix several issues on Experience Index schema

lebaudantoine:
## Purpose

Fix several issues on Experience Index schema

## Proposal

- [x] Rename protected column name 'format'
- [x] Fix inheritance issue of table args inspired by [this workaround](https://docs.sqlalchemy.org/en/14/orm/declarative_mixins.html#combining-table-mapper-arguments-from-multiple-mixins)

Please note, that SQLModel schema might be refactored at some point when switching from SQLModel to native SQLAlchemy.


src/api/core/warren/migrations/versions/94d0308156c1_fix_xi_time_constraint.py
```diff
@@ -0,0 +1,35 @@
+"""Fix xi time constraint
```
lebaudantoine, line 1: might be improved, to a more explicit description.

src/api/core/warren/migrations/versions/86ec32885672_rename_format_experience.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision: str = "86ec32885672"
+down_revision: Union[str, None] = "94d0308156c1"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column("experience", "format", new_column_name="technical_datatypes")
```
lebaudantoine, line 21: note: will not work with mysql.

jmaupetit: Why?

src/api/core/warren/migrations/versions/94d0308156c1_fix_xi_time_constraint.py
jmaupetit: Should be fine. As long as the commit message is explicit, we are fine.

jmaupetit (approved):

src/api/core/warren/migrations/versions/86ec32885672_rename_format_experience.py
lebaudantoine: [based on this discussion](https://stackoverflow.com/questions/22599391/how-to-use-alter-column-in-alembic#comment96733376_47184837)

PR #156, opened 2023-12-01
⚗️(notebook) compute warren-tdbp indicator requirements

quitterie-lcs:
## Purpose

Explore feasbility of indicator requirements for TdBP project 

## Proposal

A notebook computing floating frame, active actions and dynamic cohort based on
dataset provided by Avignon university.

### DIFF
--- a/notebook/notebooks/tdbp-indicators.md
+++ b/notebook/notebooks/tdbp-indicators.md
@@ -1,0 +1,201 @@
+---
+jupyter:
+  jupytext:
+    text_representation:
+      extension: .md
+      format_name: markdown
+      format_version: '1.3'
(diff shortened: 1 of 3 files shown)


notebook/notebooks/tdbp-indicators.md
```diff
@@ -0,0 +1,201 @@
...
+]
+
+# Get the first action in the course
+
+# TODO first_action with lrs_query
+```
+
+```python
+def compute_active_actions(
+    action_type: str, date: datetime = None, window_size: int = 15
```
quitterie-lcs, line 117: `action_type` is out of date for active actions. It should be computed on ALL course actions

PR #160, opened 2023-12-04
✅(api) add load tests for a video route

jmaupetit:
## Purpose

## Proposal

- [x] add a test to reproduce the issue
- [ ] fix the issue


jmaupetit:
### DO NOT MERGE ###

This PR is still a test and increase too much tests execution time to be worth it.

wilbrdt:
Closing as outdated and as it would increase too much tests execution time

PR #161, opened 2023-12-04
🐛(api) force single database engine and session per thread

jmaupetit:
## Purpose

## Proposal

Use the singleton pattern to avoid opening too many connections to the database and reach the maximal connection pool for database requests during API heavy load.

### DIFF
--- a/src/api/core/warren/api/__init__.py
+++ b/src/api/core/warren/api/__init__.py
@@ -1,13 +1,25 @@
 """Warren API root."""
+from contextlib import asynccontextmanager
+
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 
 from warren.conf import settings
+from warren.db import get_engine
 
 from .health import router as health_router
 from .v1 import app as v1
 
-app = FastAPI()
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    """Application life span."""
+    engine = get_engine()
+    yield
+    engine.dispose()
+
+
+app = FastAPI(lifespan=lifespan)
 
 
 app.add_middleware(
--- a/src/api/core/warren/db.py
+++ b/src/api/core/warren/db.py
@@ -1,29 +1,76 @@
 """Warren persistence database connection."""
+import logging
+from typing import Optional
 
-from typing import Generator
-
+from sqlalchemy import Engine as SAEngine
 from sqlalchemy import text
 from sqlalchemy.exc import OperationalError
-from sqlalchemy.orm import Session as SASession
-from sqlmodel import Session, create_engine
+from sqlmodel import Session as SMSession
+from sqlmodel import create_engine
 
 from .conf import settings
 
-engine = create_engine(settings.DATABASE_URL, echo=settings.DEBUG)
+logger = logging.getLogger(__name__)
 
 
-def get_session() -> Generator[Session, None, None]:
-    """Get database session single instance."""
-    with Session(engine) as session:
-        yield session
+class Singleton(type):
+    """Singleton pattern metaclass."""
+
+    _instances: dict = {}
+
+    def __call__(cls, *args, **kwargs):
+        """Store instances in a private class property."""
+        if cls not in cls._instances:
+            cls._instances[cls] = super().__call__(*args, **kwargs)
+        return cls._instances[cls]
+
+
+class Engine(metaclass=Singleton):
+    """Database engine singleton."""
+
+    _engine: Optional[SAEngine] = None
+
+    def get_engine(self, url, echo=False) -> SAEngine:
+        """Get created engine or create a new one."""
+        if self._engine is None:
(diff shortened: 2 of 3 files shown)


jmaupetit:
This is an alternative to #137 originally proposed by @lebaudantoine

src/api/core/warren/indicators/mixins.py
```diff
@@ -193,7 +195,7 @@ async def get_or_compute(self, update: bool = False):
         to_update = []
         for cache in caches:
             if isinstance(cache, CacheEntry) and not update:
-                logger.debug("Cache entry with ID %s wont be updated", str(cache.id))
+                # logger.debug("Cache entry with ID %s wont be updated", str(cache.id))
```
lebaudantoine, line 198: dead code.

src/api/core/logging-config.dev.yaml
```diff
@@ -37,6 +37,11 @@ loggers:
       - default
     level: DEBUG
     propagate: false
+  warren:
+    handlers:
+      - default
+    level: DEBUG
+    propagate: false
```
lebaudantoine, line 44: I might see why you're declaring this logger. Any collusion with the `core.warren.api.v1` logger that has it propagate set to false ?

src/api/core/warren/db.py
```diff
@@ -1,26 +1,74 @@
...
+def get_engine():
+    """Get database engine."""
+    return Engine().get_engine(url=settings.DATABASE_URL, echo=settings.DEBUG)
+
+
+def get_session():
+    """Get database session."""
+    session = Session().get_session(get_engine())
+    logger.debug("Getting session %s", session)
+    return session
```
lebaudantoine, line 63: Nice design. However, I find it a bit convoluted to have:

- A class, with a `get_session` method.
- A function `get_session`.
- In the cache, a property `db_session`.

We make three nested calls to get a database session. Wdyt?

src/api/core/warren/api/__init__.py
```diff
@@ -1,13 +1,25 @@
...
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    """Application life span."""
+    engine = get_engine()
+    yield
+    engine.dispose()
+
+
+app = FastAPI(lifespan=lifespan)
```
lebaudantoine, line 22: I see kind of a dependency issue to `FastAPI` in this case, based on some remarks of the first implementation I made to fix the session issue.

We rely on `lifespan` to properly close any open session (and connection pool through the engine).
Using your singleton and helper function without `FastAPI` or uvicorn, won't work properly. I thought you wanted to avoid such cases.

src/api/core/warren/indicators/mixins.py
jmaupetit: Nope, in fact it's a bug. I should have fixed it: I the object has been saved and not refreshed, there is no id associated with this record...

src/api/core/warren/db.py
```diff
@@ -1,26 +1,74 @@
...
+class Singleton(type):
+    """Singleton pattern metaclass."""
+
+    _instances = {}
+
+    def __call__(cls, *args, **kwargs):
+        """Store instances in a private class property."""
+        if cls not in cls._instances:
+            cls._instances[cls] = super().__call__(*args, **kwargs)
+        return cls._instances[cls]
```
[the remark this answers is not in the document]
jmaupetit: Not sure about this as it's only used here for now.

src/api/core/warren/db.py
jmaupetit: In fact, only the `get_session` and `get_engine` functions should be public...

src/api/core/warren/api/__init__.py
jmaupetit: I think we might need to document this, because, providing an utility that wraps the `engine.dispose()` method seems overkill. In a library/script usage of indicators, one might dispose the engine explicitly during the script teardown. It's SQLAlchemy business-as-usual logic. WDYT?

src/api/core/warren/indicators/mixins.py
jmaupetit: Wasn't able to reproduce the bug. I've restored the debug log message.

src/api/core/warren/api/__init__.py
lebaudantoine: I am fine with your implementation. I was just stating we still rely on fastAPI to dispose the engine.

src/api/core/warren/db.py
lebaudantoine: alright your call !

lebaudantoine (approved):
LGTM, as soon it's functional

src/api/core/logging-config.dev.yaml
jmaupetit: Removed it!

PR #170, opened 2023-12-11
🔧(api) configure FastAPI integration for sentry

quitterie-lcs:
## Purpose

Sentry needs to be integrated for `warren` API

## Proposal

Sentry is configured for warren API following this integration:
https://docs.sentry.io/platforms/python/integrations/fastapi/


.env.dist
```diff
@@ -34,6 +34,10 @@ WARREN_APP_ALLOWED_HOSTS=localhost
 WARREN_API_ROOT_URL=http://localhost:8100
 WARREN_APP_ROOT_URL=http://localhost:8090
 
+# Sentry
+WARREN_SENTRY_DSN=https://fake@key.ingest.sentry.io/1234567
+WARREN_EXECUTION_ENVIRONMENT=development
```
jmaupetit, line 39: I would not activate it by default:

```suggestion
# WARREN_SENTRY_DSN=https://fake@key.ingest.sentry.io/1234567
# WARREN_EXECUTION_ENVIRONMENT=development
```

jmaupetit (approved):
Two small fixes and we are good to go!

PR #174, opened 2023-12-11
Distribute API database migrations

jmaupetit:
## Purpose

When installing Warren API python package, one need to be able to run database-related operations (migrations) using alembic. As alembic requires a per-project configuration, we need to distribute alembic configuration in the API core python package, and provide a wrapper script that runs alembic behind the scene.

## Proposal

- [x] move alembic configuration file
- [x] add alembic configuration to python package
- [x] add warren CLI for database migrations
- [x] add tests for the command line parser

### DIFF
--- a/src/api/core/warren/commands.py
+++ b/src/api/core/warren/commands.py
@@ -1,0 +1,71 @@
+"""warren-api script."""
+
+import argparse
+import copy
+import sys
+from typing import List
+
+from warren import migrations
+
+
+def get_command_line_parser():
+    """Get command line arguments parser."""
+    parser = argparse.ArgumentParser(prog="warren")
+
+    # Migrations
+    alembic = parser.add_subparsers(title="Database migrations")
+
+    current = alembic.add_parser("current", help="display the current migration")
+    current.add_argument(
+        "-v",
+        "--verbose",
+        default=False,
+        action="store_true",
+        help="verbose mode",
+    )
+    current.set_defaults(func=migrations.current, verbose=False)
+    downgrade = alembic.add_parser(
+        "downgrade", help="downgrade to a specific migration"
+    )
+    downgrade.add_argument(
+        "revision", type=str, help="the target migration to downgrade to"
+    )
+    downgrade.set_defaults(func=migrations.downgrade)
+    history = alembic.add_parser("history", help="display migrations history")
+    history.add_argument(
+        "-v",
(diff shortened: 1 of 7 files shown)


src/api/core/warren/tests/test_commands.py
```diff
@@ -0,0 +1,129 @@
...
+    """Test warren command call without subcommands."""
+    commands.main([])
+    captured = capsys.readouterr()
+    assert captured.out.startswith(
+        "usage: warren [-h] {current,downgrade,history,upgrade}"
+    )
+
+
+def test_current_command(monkeypatch):
+    """Test warren current commands."""
```
jmaupetit, line 86:
```suggestion
    """Test warren current command."""
```

jmaupetit: ditto for the following tests

src/api/core/warren/commands.py
```diff
@@ -0,0 +1,70 @@
...
+        default="head",
+        nargs="?",
+        help="the target migration to upgrade to (default: head)",
+    )
+    upgrade.set_defaults(func=migrations.upgrade)
+
+    return parser
+
+
+def main(cmd: list[str]):
```
quitterie-lcs, line 55:
```suggestion
def main(cmd: List[str]):
```

src/api/core/warren/commands.py
```diff
@@ -0,0 +1,70 @@
+"""warren-api script."""
+
+import argparse
+import copy
+import sys
+
```
quitterie-lcs, line 6:
```suggestion

from typing import List
```

src/api/core/warren/migrations/__init__.py
```diff
@@ -0,0 +1,28 @@
+"""Database migrations helpers."""
+from pathlib import Path
```
quitterie-lcs, line 2:
```suggestion

from pathlib import Path
```

src/api/core/warren/tests/test_commands.py
```diff
@@ -0,0 +1,129 @@
+"""Test Warren commands functions."""
+from unittest.mock import MagicMock
```
quitterie-lcs, line 2:
```suggestion

from unittest.mock import MagicMock
```

quitterie-lcs (approved):

PR #177, opened 2023-12-13
🐛(api) fix warren command line parser

jmaupetit:
## Purpose

With the program name passed as the first argument from sys.argv, the argument parser does not validate the command.

## Proposal

- [x] remove the program name (if present)

### DIFF
--- a/src/api/core/warren/commands.py
+++ b/src/api/core/warren/commands.py
@@ -2,8 +2,7 @@
 
 import argparse
 import copy
-import sys
-from typing import List
+from typing import List, Optional
 
 from warren import migrations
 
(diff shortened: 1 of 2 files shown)


quitterie-lcs (approved):
Sounds good

src/api/core/warren/commands.py
```diff
@@ -53,9 +52,10 @@ def get_command_line_parser():
     return parser
 
 
-def main(cmd: List[str]):
+def main(cmd: Union[List[str], None] = None):
```
quitterie-lcs, line 55:
```suggestion
def main(cmd: Optional[List[str]] = None):
```

quitterie-lcs (commented):
Tiny improvement

PR #179, opened 2023-12-14
📦️(frontend) integrate changesets to publish npm packages

jmaupetit:
## Purpose

We need to publish frontend components we use for Warren UI so that plugins developer can use them in their own plugins.

## Proposal

As recommended by TurboJS, we've integrated the changesets tool to handle frontend related changes and packages publication to npm.


docker-compose.yml
```diff
@@ -76,6 +76,7 @@ services:
       - dev
     volumes:
       - ./src/frontend:/app
+      - .git:/app/.git
```
lebaudantoine, line 79: I am not totally sure to get this trick 🧐.

src/frontend/Dockerfile
```diff
@@ -34,6 +34,11 @@ WORKDIR /app
 # -- Development --
 FROM core as development
 
+# Install Git (required for changeset)
+RUN apt-get update && \
+    apt-get install -y git && \
+    rm -rf /var/lib/apt/lists/*
+
```
lebaudantoine, line 41: Let's discuss it, to understand this requirement!

lebaudantoine (approved):
LGTM

src/frontend/Dockerfile
jmaupetit: Git is required by changesets. If not installed the CLI fails while booting.

docker-compose.yml
jmaupetit: Git and a git history are required for the `changeset` CLI to run.

jmaupetit: That's the best I was able to do.

PR #189, opened 2024-01-16
📝(project) improve mike local usage

jmaupetit:
## Purpose

Our CI bot should not be the only user allowed to act on the documentation. If we need to perform clean up actions from our workstation, Mike needs to be able to commit on Warren's repository on our behalf.

## Proposal

- [x] create a `docs` user in the docs Docker image
- [x] share local SSH and Git configuration with the Docker user so that Mike can act on the remote repository


wilbrdt (approved):
LGTM! :tada:

PR #191, opened 2024-01-17
🐛(docs) fix Docker image build for MacOS users

jmaupetit:
## Purpose

MacOS users default group ID is `20` (staff). This group often already exists on GNU/Linux distributions, hence we should not enforce its creation if it already exists, or else the `docs` Docker image build fails.

## Proposal

Make group creation optional.


lebaudantoine (approved):
Tested and approved!

PR #193, opened 2024-01-21
🧑‍💻(project) introduce a clean command

lebaudantoine:
Allow developer cleaning their repo, by restoring it as it was freshly cloned. Especially useful to solve recurrent out-dated env files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -388,6 +388,10 @@ help:
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 .PHONY: help
 
+clean: ## restore repository state as it was freshly cloned
+	git clean -idx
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -388,6 +388,10 @@ help:
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 .PHONY: help
 
+clean: ## restore repository state as it was freshly cloned
+	git clean -idx
```
jmaupetit, line 392: Some extremists would propose the following:

```suggestion
	git clean -fdx
```

But your moderate suggestion is perfectly fine :sweat_smile:

jmaupetit (approved):

Makefile
lebaudantoine: I buy it !

jmaupetit: Will merge this PR as is. We'll see if we must be more extremist or not! :smiley_cat:

PR #203, opened 2024-01-31
📝(api) add CLI documentation

jmaupetit:
## Proposal

Automate CLI documentation generation using mkdocs-click.


src/api/core/warren/conf.py
quitterie-lcs, line 1: I would add default value for settings in a dedicated commit. WDYT?

PR #204, opened 2024-02-01
Add XI CLI commands

jmaupetit:
## Purpose

We need a CLI command to (list and) index Moodle courses and content in the Experience Index.

## Proposal

- [x] add `xi list` commands
- [x] add `xi index` commands
- [x] write tests
- [x] add explicit error message for moodle login issues
- [x] ~xi routes should be protected behind authentication~ postponed this in a new PR

### DIFF
--- a/src/api/core/warren/cli.py
+++ b/src/api/core/warren/cli.py
@@ -6,11 +6,19 @@ import logging
 import sys
 from inspect import Parameter, Signature, signature
 from typing import Optional
+from uuid import UUID
 
 import click
 from pydantic import BaseModel
 
 from warren import __version__ as warren_version
+from warren.xi.client import ExperienceIndex
+from warren.xi.enums import AggregationLevel
+from warren.xi.indexers.moodle.client import Moodle
+from warren.xi.indexers.moodle.etl import (
+    CourseContent,
+    Courses,
+)
 
 from . import migrations as alembic_migrations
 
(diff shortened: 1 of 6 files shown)


src/api/core/warren/xi/client.py
```diff
@@ -109,7 +109,7 @@ async def get(self, object_id: Union[UUID, IRI]) -> Union[ExperienceRead, None]:
             return None
 
         response.raise_for_status()
-        return ExperienceRead.construct(**response.json())
+        return ExperienceRead(**response.json())
```
quitterie-lcs, line 112: Why do you validate the response ?

jmaupetit: Pydantic's models `construct` method does not parse nested models for performance reasons. If we want a complete `ExperienceRead` instance while getting an experience from the XI, we need to parse AND validate API
response. This decreases performances but has a reasonable trade-off.

src/api/core/warren/cli.py
```diff
@@ -176,3 +184,129 @@ def indicator_compute(ctx: click.Context, indicator: str, cache: bool):
     click.echo(
         result.json() if issubclass(compute_annotation, BaseModel) else str(result)
     )
+
+
+# -- EXPERIENCE INDEX (AKA XI) COMMAND --
```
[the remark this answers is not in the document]
jmaupetit: There's only one `xi` command but multiple sub-commands. The same applies for many commands. :wink:

wilbrdt (commented):
Alt text
Looks good to me! Only fixing some typos

Note: don't forget to check coverage of `cli.py` (a test is probably missing for the `xi list content` command)

jmaupetit:
> Note: don't forget to check coverage of `cli.py` (a test is probably missing for the `xi list content` command)

Well spotted! I've added a test for this.

wilbrdt (approved):

PR #207, opened 2024-02-08
Fix Ralph 4 integration

jmaupetit:
## Purpose

We recently switched the project to Ralph 4 and some configuration are not fully compatible.

## Proposal

- [x] fix Ralph authentication file creation
- [x] fix `make fixtures` command
- [x] improve ES integration


wilbrdt (approved):
Looks good to me!

PR #211, opened 2024-02-21
♻️(api) configure root path global variable for alembic configuration

quitterie-lcs:
## Purpose

Alembic config file is embedded in Warren lib packaging, however, it needs a
relative path to be accessed. Otherwise a redefinition is necessary.

## Proposal

Add a `ROOT_PATH` setting variable for alembic config file relative path
location.

### DIFF
--- a/src/api/core/warren/conf.py
+++ b/src/api/core/warren/conf.py
@@ -43,6 +43,13 @@ class Settings(BaseSettings):
         "actor.openid",
     }
 
+    # API Core Root path
+    # (used at least by everything that is alembic-configuration-related)
+    ROOT_PATH: Path = Path(__file__).parent
+
+    # Alembic
+    ALEMBIC_CFG_PATH: Path = ROOT_PATH / "alembic.ini"
+
     # Security
     ALLOWED_HOSTS: List[str] = [
         "http://localhost:8090",
--- a/src/api/core/warren/migrations/__init__.py
+++ b/src/api/core/warren/migrations/__init__.py
@@ -1,12 +1,11 @@
 """Database migrations helpers."""
 
-from pathlib import Path
-
 from alembic import command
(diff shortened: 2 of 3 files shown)


src/api/core/warren/conf.py
```diff
@@ -43,6 +43,9 @@ class Settings(BaseSettings):
         "actor.openid",
     }
 
+    # API migration
```
jmaupetit, line 46:
```suggestion
    # API Core Root path
    # (used at least by everything that is alembic-configuration-related)
```

src/api/core/warren/migrations/__init__.py
```diff
@@ -1,12 +1,11 @@
...
-from pathlib import Path
-
 from alembic import command
 from alembic.config import Config
 
-ROOT_PATH: Path = Path(__file__).parent.parent
-ALEMBIC_CFG: Config = Config(ROOT_PATH / "alembic.ini")
+from ..conf import settings
+
+ALEMBIC_CFG: Config = Config(settings.ROOT_PATH / "alembic.ini")
```
jmaupetit, line 8: Maybe this should also be a global setting?

src/api/core/warren/tests/fixtures/db.py
```diff
@@ -20,7 +20,7 @@ def db_engine():
     SQLModel.metadata.create_all(engine)
 
     # Pretend to have all migrations applied
-    alembic_cfg = Config("core/warren/alembic.ini")
+    alembic_cfg = Config(settings.ROOT_PATH / "alembic.ini")
```
jmaupetit, line 23: Hence, we can use it as follow:

```suggestion
    alembic_cfg = settings.ALEMBIC_CFG

jmaupetit (approved):

PR #212, opened 2024-02-21
Improve XI API and client

jmaupetit:
## Purpose

A suite of various improvements around the XI.

## Proposal

- [x] fix missing xi routers DB rollbacks
- [x] make XI client more developer-friendly
- [x] use the XI client in the seeding script

### DIFF
--- a/src/api/core/warren/xi/client.py
+++ b/src/api/core/warren/xi/client.py
@@ -1,14 +1,16 @@
 """Experience Index HTTP client."""
 
+import json
+import logging
 from abc import ABC, abstractmethod
 from http import HTTPStatus
-from os.path import join
+from pathlib import PurePath
 from typing import List, Optional, Protocol, Tuple, Union
 from urllib.parse import quote_plus
 from uuid import UUID
 
 from httpx import AsyncClient
-from pydantic import parse_obj_as
+from pydantic import UUID4, parse_obj_as
 
 from warren.conf import settings
 from warren.fields import IRI
@@ -27,10 +29,8 @@ from .models import (
     RelationRead,
     RelationUpdate,
 )
-from .schema import (
-    Experience,
-    Relation,
-)
+
+logger = logging.getLogger(__name__)
 
 
 class Client(Protocol):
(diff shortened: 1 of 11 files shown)


src/api/core/warren/xi/client.py
```diff
@@ -190,22 +215,39 @@ async def delete(self, **kwargs) -> None:
...
+        db_relations = await self.read(
+            source=data.source_id, target=data.target_id, kind=data.kind.value
+        )
+
+        # Create
+        if db_relations is None or len(db_relations) == 0:
+            return await self.create(data)
+
+        # Update
+        db_relation = db_relations.pop()
```
wilbrdt, line 229: Could we have multiple relations objects with the same source, target and kind?
Should we check if there are multiple relations objects before updating the last read?

src/api/core/warren/xi/routers/experiences.py
```diff
@@ -64,6 +66,10 @@ async def read_experiences(  # noqa: PLR0913
         statement = statement.where(Experience.aggregation_level == aggregation_level)
     if iri:
         statement = statement.where(Experience.iri == iri)
+    if technical_datatypes:
+        statement = statement.filter(
+            Experience.technical_datatypes.comparator.contains(technical_datatypes)  # type: ignore[union-attr]
```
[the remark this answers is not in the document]
jmaupetit: Yup, that's it :point_right: https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.JSON.Comparator

src/api/core/warren/xi/client.py
jmaupetit: No we can't: we have a unique constraint on the `("source_id", "target_id", "kind")` triplet.

Reference: https://github.com/openfun/warren/blob/main/src/api/core/warren/xi/schema.py#L49

wilbrdt: Great! No problem here then

wilbrdt (approved):

PR #213, opened 2024-02-23
🔒️(lti) change lti user identifying parameter

wilbrdt:
## Purpose

Moodle has mapped the `lis_person_sourcedid` parameter to its optionnal ID
number, often not specified.

## Proposal

As Warren does not need to have a username but only a unique identifier for
each user, it can use the opaque `user_id` parameter along with the consumer
site url.

On a side note, adding some logs to ease debugging.


quitterie-lcs (approved):
Looks good to me

src/api/core/warren/db.py
```diff
@@ -72,6 +72,7 @@ def is_alive() -> bool:
     try:
         session.execute(text("SELECT 1 as is_alive"))
         return True
-    except OperationalError:
+    except OperationalError as err:
+        logger.debug("Exception: %s", err)
```
jmaupetit, line 76: Sure we need this in this PR?

jmaupetit: Okay then.

jmaupetit (approved):

PR #215, opened 2024-02-26
🔧(helm) multiple changes for deploying with `stack` 

wilbrdt:
### Naming convention

- [x] Some templates follows the snake_case naming convention. Renaming them to kebab-case.

### Secrets and configmaps

- [x] Now that we use Hashicorp Vault for managing secrets, we can now create secrets to store sensitive environment variables within the chart. Additionally, non-sensitive environment variables are now appropriately segregated withing configmaps (api/app env, database and logging).

### Minor lint issue

- [x] Fixing the lint issue by only adding an `httpHeaders` if value `allowHosts` is not empty, for both `api` and `app` deployments.

### Migration and staticfiles collection jobs

- [x] Using pre-install Helm hooks for migrating the database and collecting static files before a deployment. Another solution would have been an `initContainer` inside the deployment, either executing the migration/collection, or waiting for jobs doing the migration/collection.


wilbrdt:
Closing as work is done on #252

PR #217, opened 2024-02-29
🐛(cli) enforce key-value argument splitting to 1

quitterie-lcs:
## Purpose

By default, split() built-in method splits on all occurences of separator. For
the argument key-value splitting in CLI, some values can be IRIs which contains
an equal sign.

## Proposal

 To secure the splitting with IRI, we enforce the splitting to 1
occurence.


wilbrdt (approved):
Looks good to me! :tada:

PR #218, opened 2024-02-29
🐛(frontend) add missing CSS files in packages

jmaupetit:
## Purpose

We need to embed Scss compiled styles to frontend packages so that all components can benefit from them.

## Proposal

- [x] integrate the `esbuild-sass-plugin` with `tsup`

### DIFF
--- a/src/frontend/packages/core/.eslintignore
+++ b/src/frontend/packages/core/.eslintignore
@@ -1,1 +1,2 @@
 node_modules
+tsup.config.ts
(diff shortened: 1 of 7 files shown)


jmaupetit:
This fix is inspired from this comment: https://github.com/vercel/turbo/discussions/692#discussioncomment-6415071

jmaupetit:
Current work addresses the same issue as #216

lebaudantoine (approved):

PR #221, opened 2024-03-02
👷(circle) add a helm chart publication job

wilbrdt:
## Purpose

It would be nice that CI automatically published Warren Helm chart package when a `vX.X.X-helm` tag has been pushed.

## Proposal

- [x] Adding a helm chart publication job. Note that Helm chart is not tested by the CI prior to publication, and should be tested manually before tagging.
- [x] Updating documentation on release git tag pattern.


quitterie-lcs (approved):
Sounds good !

PR #223, opened 2024-03-04
🔧(app) add global settings for LTI parameters

quitterie-lcs:
## Purpose

LTI parameters for development and tests are locally hard coded.

## Proposal

We add global settings in Django configuration to be able to overload settings value in Warren
package development context.


src/app/warren/settings.py
```diff
@@ -341,6 +342,25 @@ class Development(Base):
     DEBUG = True
     ALLOWED_HOSTS = ["*"]
 
+    # LTI parameters
+    LTI_PARAMETERS = {
+        "lti_message_type": "basic-lti-launch-request",
+        "lti_version": "LTI-1p0",
+        "resource_link_id": str(uuid.uuid4()),
+        "lis_person_contact_email_primary": Value(
+            "johndoe@example.com", environ_name="LIS_PERSON_CONTACT_EMAIL_PRIMARY"
```
lunika, line 351: I don't know how settings are managed in warren but there is no need to specify the prefix here ? I see `WARREN` is used in you dist file.

lunika: The answer is here https://github.com/openfun/warren/blob/39205a98092ac599046362bc035fca27cd43c885/src/app/warren/settings.py#L36C1-L36C23

src/app/warren/settings.py
```diff
@@ -341,6 +342,25 @@ class Development(Base):
...
+        ),
+        "user_id": Value("1234", environ_name="USER_ID"),
```
lunika, line 353:
```suggestion
        "user_id": Value("1234", environ_name="LTI_USER_ID"),
```

I suggest you to use `LTI_USER_ID`. `USER_ID` is really generic and you can probably have conflicts.

src/app/warren/settings.py
```diff
@@ -341,6 +342,25 @@ class Development(Base):
...
+        "context_id": Value(
+            "course-v1:openfun+mathematics101+session01", environ_name="LTI_CONTEXT_ID"
+        ),
+        "context_title": Value("Mathematics 101", environ_name="LTI_CONTEXT_TITLE"),
+        "roles": Value("student", environ_name="LTI_ROLES"),
+        "launch_presentation_locale": Value(
+            "fr", environ_name="LAUNCH_PRESENTATION_LOCALE"
```
lunika, line 360:
```suggestion
            "fr", environ_name="LTI_LAUNCH_PRESENTATION_LOCALE"
```

I think ahving `LTI_` on every settings is more consistent.

src/app/warren/settings.py
quitterie-lcs: Yup!

lunika (approved):

PR #245, opened 2024-05-03
✨(api) change cache key to instance attributes

wilbrdt:
## Purpose

As of now, an indicator may return different results for the same LRS query (for example with a boolean flag passed as an attribute, or an env setting), we cannot use the cache mixins, as we only rely on the LRS query as a cache key.

## Proposal

Changing it for a hash from its instance attributes, so cache mixins are usable for these indicators.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to
 ### Changed
 
 - Restrict dashboards access to instructors and administrators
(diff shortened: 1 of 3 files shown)


quitterie-lcs (commented):
You can also adapt the `get_lrs_query()` method in other tests where an lrs query instance is still returned. `pass` instruction should fit for all the tests

lunika (approved):

PR #251, opened 2024-05-16
✨(cli) add migration check command

wilbrdt:
## Purpose

The Warren CLI Alembic wrapper lacked a command to verify if the database has any unapplied migrations.                                                       

## Proposal

This update introduces a migration check command, which will exit with a status of 1 if there are unapplied migrations in the target database.

### DIFF
--- a/src/api/core/warren/cli.py
+++ b/src/api/core/warren/cli.py
@@ -9,6 +9,7 @@ from typing import Optional
 from uuid import UUID
 
 import click
+from alembic.util import CommandError
 from pydantic import BaseModel
 
 from warren import __version__ as warren_version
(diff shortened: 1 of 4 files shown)


src/api/core/warren/cli.py
```diff
@@ -42,6 +43,15 @@ def migration():
     """Database migration commands (alembic wrapper)."""
 
 
+@migration.command()
+def check():
```
quitterie-lcs, line 47: Maybe can you display a message when all migrations are applied? WDYT?

src/api/core/warren/cli.py
```diff
@@ -42,6 +43,15 @@ def migration():
...
+    """Show current database migration."""
```
lunika, line 48:
```suggestion
    """check current database migration."""
```

lunika (approved):

PR #255, opened 2024-05-22
Refacto for dashboard depending on LTI roles

quitterie-lcs:
## Purpose

Support dedicated views for dashboards according to LTI user roles

## Proposal

Add role in appData returned from LTIView and propose different views in the frontend depending on the LTI user role returned by the app


src/app/apps/lti/views.py
```diff
@@ -136,6 +132,7 @@ def _do_on_success(self, lti_request: LTI, *args, **kwargs) -> HttpResponse:
 
         self.app_data = {
             "lti_route": kwargs["selection"] or "demo",
+            "roles": lti_request.roles,
```
lunika, line 135: I would replace this by something like `is_instructor`. Having to deal only with a boolean in the front application will ease its usage. Otherwise everywhere in the front application you will have to check for several roles. here a simple `context.is_instructor` will give you the information you want.

quitterie-lcs: Indeed, that's a good idea !

src/app/apps/lti/views.py
```diff
@@ -136,6 +132,7 @@ def _do_on_success(self, lti_request: LTI, *args, **kwargs) -> HttpResponse:
 
         self.app_data = {
             "lti_route": kwargs["selection"] or "demo",
+            "is_instructor": True if lti_request.is_instructor else False,
```
lunika, line 135:
```suggestion
            "is_instructor": lti_request.is_instructor,
```

lunika (approved):

PR #258, opened 2024-05-28
✨(xi) add a xi command to index all courses and their content

wilbrdt:
## Proposal

Adding a useful `warren xi index all` command to index all courses of a LMS
instance, and their content.

### DIFF
--- a/src/api/core/warren/cli.py
+++ b/src/api/core/warren/cli.py
@@ -314,9 +314,53 @@ def xi_index_course_content(
     moodle_ws_token: str,
     ignore_errors: bool,
 ):
-    """Index LMS course content."""
+    """Index LMS content of a course."""
     asyncio.run(
         _xi_index_course_content(
             course_id, xi_url, moodle_url, moodle_ws_token, ignore_errors
         )
     )
+
+
+async def _xi_index_all(
+    xi_url: str,
+    moodle_url: str,
+    moodle_ws_token: str,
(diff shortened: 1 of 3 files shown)


src/api/core/warren/cli.py
```diff
@@ -314,9 +314,53 @@ def xi_index_course_content(
...
+    await indexer_courses.execute()
+
+    experiences = await xi.experience.read(aggregation_level=AggregationLevel.THREE)
+    for experience in experiences:
+        course = await xi.experience.get(object_id=experience.id)
+        if course is None:
+            raise click.BadParameter(
+                f"Unknown course {course}. Course indexation has failed!"
+            )
+        indexer_content = CourseContent(
```
quitterie-lcs, line 351: What happened if in the experiences read in level THREE, we have non Moodle courses? (I am aware that it is not the case for now :wink:). Could be nice to handle such case now in the implementation, to avoid facing an issue later ?

quitterie-lcs: I agree to dedicate another PR to this

quitterie-lcs (approved):
Sounds good, but we should keep in mind improvements for later

src/api/core/warren/cli.py
wilbrdt: Opened issue https://github.com/openfun/warren/issues/260

PR #259, opened 2024-05-31
🔒️(project) securely pass user and course ids through jwt

wilbrdt:
## Purpose

Previously, `course_id` was transmitted to the frontend through app context data.
`user_id` was not transmitted to the frontend.

## Proposal

To be secure, both `course_id` and `user_id` will now be transmitted through the LTI token instead of the app context data.
This change prevents users from altering the course_id or user_id to access unauthorized data.

### DIFF
--- a/src/app/apps/lti/views.py
+++ b/src/app/apps/lti/views.py
@@ -22,7 +22,7 @@ from lti_toolbox.views import BaseLTIView
 from oauthlib import oauth1
 from rest_framework_simplejwt.exceptions import TokenError
 
-from .forms import BaseLTIUserForm
+from .forms import BaseLTIContextForm, BaseLTIUserForm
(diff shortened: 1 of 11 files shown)


src/app/apps/lti/views.py
```diff
@@ -31,18 +31,36 @@
...
+            "course_id": (
+                lti_request.origin_url
+                if lti_request.is_moodle_format
+                else lti_request.get_param("context_id")
+            ),
+        }
+
+        lti_context_form = BaseLTIContextForm(lti_context_data)
+        if not lti_context_form.is_valid():
+            logger.debug("LTI context is not valid: %s", lti_context_form.errors)
```
lunika, line 51: Why a debug level ? It is only neeed in a debug context ?

lunika (approved):

quitterie-lcs (approved):
LGTM !

PR #268, opened 2024-06-24
Decode and store JWT token in frontend

quitterie-lcs:
## Purpose

Jwt token contains course and user information, that can be accessed after
decoding the token. As a decoding step is time consuming, it is executed once
and the content is available for the whole app.

## Proposal

`decodeJwtLTI` utils is now available in warren-core to decode the JWT token and
use `course_id`, `user` or user `roles` everywhere in the application.

### DIFF
--- a/src/frontend/packages/core/src/hooks/useRefreshToken.ts
+++ b/src/frontend/packages/core/src/hooks/useRefreshToken.ts
@@ -1,9 +1,12 @@
 import { AxiosInstance } from "axios";
 import { AppData } from "../types";
+import { decodeJwtLTI } from "../utils";
 import { useLTIContext } from "./useLTIContext";
+import { useJwtContext } from "./useJwtContext";
 
 export const useRefreshToken = (client: AxiosInstance) => {
   const { appData, setAppData } = useLTIContext();
+  const { setDecodedJwt } = useJwtContext();
 
   const refreshAccessToken = async () => {
     const response = await client.post("token/refresh/", {
@@ -17,6 +20,8 @@ export const useRefreshToken = (client: AxiosInstance) => {
       ...prevData,
       access: newAccessToken,
     }));
+    // Decode the new access token as the JWT token
+    setDecodedJwt(decodeJwtLTI(newAccessToken));
 
     return newAccessToken;
   };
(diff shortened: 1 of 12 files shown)


src/frontend/packages/core/src/hooks/useRefreshToken.ts
```diff
@@ -1,9 +1,13 @@
 import { AxiosInstance } from "axios";
 import { AppData } from "../types";
 import { useLTIContext } from "./useLTIContext";
+import { useJwtContext } from "./useJwtContext";
+import { decodeJwtLTI
 
+ } from "../utils";
```
wilbrdt, line 7:
```suggestion
import { decodeJwtLTI } from "../utils";
```

src/frontend/packages/core/src/types/index.ts
```diff
@@ -21,3 +21,23 @@ export interface AppData {
...
+}
+
+export interface DecodedJwtLTI {
+  consumer_site: string;
+  course_id: string;
+  exp: number;
+  iat: number;
+  jti: string;
+  locale: string;
+  resource_link_description: string;
```
wilbrdt, line 37:
```suggestion
  resource_link_description?: string;
```

src/frontend/packages/core/src/utils/jwtDecode.ts
```diff
@@ -0,0 +1,30 @@
+import { jwtDecode } from "jwt-decode";
+import { DecodedJwtLTI } from "../types";
+
+export const isDecodedJwtLTI = (jwt: unknown): jwt is DecodedJwtLTI => {
+  if (jwt && typeof jwt === "object") {
+    const courseId = (jwt as DecodedJwtLTI).course_id;
+    const userId = (jwt as DecodedJwtLTI).user?.id;
```
wilbrdt, line 7: Could `user` not be set?

src/frontend/packages/core/src/utils/jwtDecode.ts
wilbrdt, line 1: It would be nice for the file to have the name of its main function `decodeJwtLTI`. WDYT?

wilbrdt (commented):
Nice work! :raised_hands:

src/frontend/packages/core/src/utils/jwtDecode.ts
quitterie-lcs: I don't think so with LTI. I'll fix it

wilbrdt (approved):

PR #283, opened 2024-09-30
♻️(indicators) move daily indicator classes to project core

quitterie-lcs:
## Purpose

Indicators used for daily computation in video plugins are generic enough to be
used in  other plugins.

## Proposal 

 Daily indicator calculation classes are moved from the dedicated plugin to the
core of the project. These classes are now accessible for reuse across other
plugins as mixins, as they provide generic functionality. Updated all relevant imports
and references to reflect the new structure.


wilbrdt (approved):
LGTM :tada:

